PR #34, opened 2023-07-25
feat: support environment variables for config

Sparkier:


backend/zeno_backend/database/database.py
```diff
@@ -164,13 +166,21 @@ def config(
...
+            return db
         else:
-            raise Exception(f"Section {section} not found in the {filename} file")
-        return db
+            db: Dict[str, Any] = {}
+            db["host"] = os.environ["DB_HOST"]
+            db["dbname"] = os.environ["DB_NAME"]
+            db["user"] = os.environ["DB_USER"]
+            db["password"] = os.environ["DB_PASSWORD"]
+            return db
```
Sparkier, line 186: This checks if there is a `database.ini`, if not, it will use env variables to configure the connection.

frontend/.env.example
```diff
@@ -0,0 +1 @@
+PUBLIC_BACKEND_ENDPOINT = "the url and port at which your backend is running, this can point to a deployed server if you only want to run the frontend"
```
Sparkier, line 1: In dev mode, this file is used. In prod, this should be the same as `process.env`.

PR #54, opened 2023-08-04
feat: implement initial version of zeno api for adding data to the ba…

Sparkier:
…ckend


Sparkier:
Sorry for the bigger change but this enables data upload to the ZenoHub backend. I tested this with an existing chatbot css that I got from an older version of Zeno using `zeno.df.to_csv('~/Downloads/data.csv')`.

Then, I used this code to upload the data which I ran from within the zeno_api folder:
```
"""Uploading existing zeno chatbot project to a zeno backend using the Zeno API."""
import argparse
import ast
import json
from pathlib import Path

import pandas as pd
from pandas.api.types import (
    is_bool_dtype,
    is_datetime64_any_dtype,
    is_numeric_dtype,
    is_string_dtype,
)

from zeno_api.api import (
    MetadataType,
    add_label,
    add_output,
    add_postdistill,
    add_predistill,
    create_project,
    upload_datapoint,
)
from zeno_api.cognito import get_user

parser = argparse.ArgumentParser(
    prog="Upload OpenAI Chat data.",
    description="Uploads data for an OpenAI Chat report that has been saved as a csv "
    "to a new Zeno project.",
)
parser.add_argument("--csv_path")
parser.add_argument("--connex_username")
parser.add_argument("--connex_password")
args = parser.parse_args()


def resolve_column_type(df: pd.DataFrame, col: str) -> MetadataType:
    """Get the Zeno MetadataType of a pandas dataframe column.

    Args:
        df (pd.DataFrame): dataframe for which to get the column type.
        col (str): name of the column for which to get the type.

    Returns:
        MetadataType: the metadata type of the column.
    """
    if is_bool_dtype(df[col]):
        return MetadataType.BOOLEAN
    if is_datetime64_any_dtype(df[col]):
        return MetadataType.DATETIME
    if is_string_dtype(df[col]):
        return MetadataType.NOMINAL
    if is_numeric_dtype(df[col]):
        return MetadataType.CONTINUOUS
    return MetadataType.OTHER


user = get_user(args.connex_username, args.connex_password)
uuid = create_project("chatbot", view="openai-chat", user=user)

data_frame = pd.read_csv(args.csv_path)
data_frame = data_frame.fillna("")
output_cols = list(
    filter(lambda col: str(col).startswith("OUTPUToutput"), data_frame.columns)
)
models = list(
    set(list(map(lambda col: str(col).replace("OUTPUToutput", ""), output_cols)))
)
predistill_cols = list(
    filter(lambda col: str(col).startswith("PREDISTILL"), data_frame.columns)
)
postdistill_cols = list(
    filter(lambda col: str(col).startswith("POSTDISTILL"), data_frame.columns)
)

file_path = Path("data.json")
for index, row in data_frame.iterrows():
    with open(file_path, "w") as f:
        messages_obj = ast.literal_eval(row["messages"])
        json.dump(messages_obj, f)
    datapoint_name = f"{index}.json"
    upload_datapoint(uuid, file_path, datapoint_name, user)
    add_label(uuid, datapoint_name, row["label"], user)
    for col in output_cols:
        model_name = str(col).replace("OUTPUToutput", "")
        add_output(uuid, datapoint_name, model_name, row[str(col)], user)
    for col in predistill_cols:
        col_name = str(col).replace("PREDISTILL", "")
        type = resolve_column_type(data_frame, str(col))
        add_predistill(uuid, datapoint_name, col_name, row[str(col)], type, user)
    for col in postdistill_cols:
        col_name = str(col).replace("POSTDISTILL", "")
        type = resolve_column_type(data_frame, str(col))
        model = ""
        for m in models:
            if col_name.endswith(m):
                model = m
                col_name = col_name.replace(m, "", 1)
        add_postdistill(
            uuid, datapoint_name, col_name, model, row[str(col)], type, user
        )

file_path.unlink()
```

The result looks something like this:
image

Sparkier:
One remaining question that I have is whether or not zeno_api should actually be in this repo or in a separate one. Most people likely only interact with api and don't need all the other code. But for this group we could also just point to the python package of zeno_api. LMK what you think.

Sparkier:
Also, I added some documentation with Sphinx, but since I'm not 100% sure where the API should live, I'm also not sure about where to host the docs.

cabreraalex:
This is EPIC

> One remaining question that I have is whether or not zeno_api should actually be in this repo or in a separate one. Most people likely only interact with api and don't need all the other code. But for this group we could also just point to the python package of zeno_api. LMK what you think.

I think it should definitely be its own repository, probably `zeno-python-client` or something (in case we have clients for other languages in the future, e.g. TS). This will be its own PyPI package people install instead of the main `zenoml` library.

> Also, I added some documentation with Sphinx, but since I'm not 100% sure where the API should live, I'm also not sure about where to host the docs.

Yeah picking a home for docs is hard. The main website is using Docusaurus currently but it doesn't play super well with automatic Python docs...

Sparkier:
> Yeah picking a home for docs is hard. The main website is using Docusaurus currently but it doesn't play super well with automatic Python docs...

I think if we do own repo for this (which I think is preferable as well), we can just host the sphinx docs on the GitHub page and link to it in the README.

cabreraalex:
This looks awesome for now. I think it might be worth thinking a bit about the API design, maybe take inspiration from WandB. They have you first authenticate, then get a "run" object which you run commands on.

For us you might do:

```
zeno.login(...)

chatbot_task = zeno.task("name", { options })

chatbot_task.upload_data(...)

chatbot_task.upload_results(...)

chatbot_task.upload_features(...)
```

Sparkier:
> This looks awesome for now. I think it might be worth thinking a bit about the API design, maybe take inspiration from WandB: https://docs.wandb.ai/quickstart#putting-it-all-together. They have you first authenticate, then get a "run" object which you run commands on.
> 
> We probably want to use an API key instead of login, but otherwise could do something vaguely similar.
> 
> For us you might do:
> 
> ```
> zeno.authenticate(...)
> 
> chatbot_task = zeno.task("name", { options })
> 
> chatbot_task.upload_data(...)
> 
> chatbot_task.upload_results(...)
> 
> chatbot_task.upload_features(...)
> ```

Totally makes sense. Can do that in this PR or get this in to have something to work with and do another one to refactor like you said (in both cases won't be able to do today).

cabreraalex:
SGTM!

PR #74, opened 2023-08-17
feat: update styles of start pages

cabreraalex:
Updated most of the start screens (login, register, etc.) to look something like this:

Screenshot 2023-08-17 at 15 17 56


Sparkier (approved):
Love it.

PR #78, opened 2023-08-19
feat: schema changes for features

cabreraalex:
Some backend schema changes. Mostly:

* Combine `metadata` `predistill` and `postdistill` into a single feature type. 
* Rename project and add a few fields.


backend/init.sql
```diff
@@ -5,12 +5,11 @@ CREATE TABLE users (
...
     uuid text PRIMARY KEY,
+    name text NOT NULL UNIQUE
     view text NOT NULL,
-    calculate_histogram_metrics boolean DEFAULT true,
+    calculate_histogram_metrics boolean NOT NULL DEFAULT FALSE
     num_items integer NOT NULL DEFAULT 10,
-    name text NOT NULL,
-    public boolean NOT NULL DEFAULT false
-);
+    public boolean NOT NULL DEFAULT false);
```
Sparkier, line 12: Nit: should either use FALSE or false consistently.

Sparkier:
Should also check if we use the column types somewhere in the frontend to make rendering decisions.

Sparkier (approved):

PR #79, opened 2023-08-20
feat: change binary metadata color if no metric

Sparkier:
If no metric is selected, the binary metadata background color was black. This has been changed to our primary color to make it match all other histogams.


Sparkier:
Before:
image

After:
image

cabreraalex (approved):

PR #80, opened 2023-08-20
fix: small fixes to chat view to make it look like a chat bubble again

Sparkier:


Sparkier:
image

cabreraalex:
Ah I didn't realize we moved the views here. Should we keep the separate instance view & loading to let people write their own views without PRs to Zeno? Or is there still a way to use local views?

cabreraalex (approved):

PR #93, opened 2023-08-22
feat: python sdk updates and minimal example

cabreraalex:


backend/zeno_backend/database/insert.py
```diff
@@ -38,11 +39,20 @@ def project(project_config: Project, user: User):
...
     db = Database()
     try:
+        # Check if project already exists in table, if so, return UUID.
+        existing_project = db.execute(
+            "SELECT uuid FROM (SELECT project_uuid FROM user_project"
+            + "WHERE user_id = %s) WHERE name = %s;",
+            [user.id, project_config.name],
+        )
+        if existing_project is not None:
+            return existing_project[0][0]
```
Sparkier, line 49: I don't think we should do this. One user can have access to two projects with the same name.

Sparkier:
Can now support URLs, locally uploaded files, and strings as data.
image
image

Sparkier (approved):

PR #100, opened 2023-08-24
refactor: update database calls to use 'with'

cabreraalex:
# Description

Changes DB calls to use with for clarity/ease

### DIFF
--- a/backend/zeno_backend/database/database.py
+++ b/backend/zeno_backend/database/database.py
@@ -185,3 +185,21 @@ class Database:
             db["user"] = os.environ["DB_USER"]
             db["password"] = os.environ["DB_PASSWORD"]
             return db
+
+    def __enter__(self):
+        """Connect to the database.
+
+        Returns:
+            Database: The database object.
+        """
+        self.connect()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        """Disconnect from the database."""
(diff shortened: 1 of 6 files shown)


Sparkier (approved):
This is great.

Any thoughts about the `connect_execute...` functions? I having them instead of having to use `with` but if you would use them inside `with` things might break.

Sparkier:
`connect_execute` will connect, execute, commit, and disconnect. Would that not interfere with `with`? I agree we want some like these just bringing up that mixing them might be problematic.

cabreraalex:
Oh no I think its fine. If you just call a regular function on the class it doesn't touch the __enter__ or __exit__ methods, and if you use with it doesn't call any other functions unless its explicit, so should be fine to have both!

Sparkier:
Are you sure? Wouldn't the last statement break here since `connect_execute` disconnects:

```
with Database() as db:
    db.execute(...)
    db.connect_execute(...)
    db.execute(...)
```

cabreraalex:
OH I see what you mean. Yeah you should not run connect_execute inside a with.... are we doing that? Or are you saying we should explicitly check for that or something?

Sparkier:
No we're not just wanted to bring up this might be an issue. Don't know what to do about it, though.

PR #101, opened 2023-08-24
fix: configure project customization visibility

Sparkier:
# Description

Configures what certain users can see for projects. This includes:

1. Only owners can administer projects.
3. Only project editors can add slices, tags, or charts.
4. 3. Chart view only visible if user can add charts (is editor) or there are already charts.


cabreraalex (approved):

Sparkier:
Hmm good point, forgot about that. Should not show the edit button I agree.

PR #104, opened 2023-08-24
fix: remove key in compare table

Sparkier:
# Description
This somehow led to a duplicate keys exception. We might want to bring this back after further investigation.


cabreraalex:
Yeah this seems like an issue we should figure out...

cabreraalex (approved):

PR #106, opened 2023-08-25
fix: Fix repeat projects, tag size display, etc.

cabreraalex:
# Description

Fix some frontend bugs such


frontend/src/routes/(app)/project/[owner]/[project]/+layout.svelte
```diff
@@ -14,19 +14,22 @@
...
-		rowsPerPage.set(data.project.samplesPerPage ?? 5);
-		slices.set(data.slices);
-		columns.set(data.columns);
-		models.set(data.models);
-		metrics.set(data.metrics);
-		folders.set(data.folders);
-		tags.set(data.tags);
+	$: setupProject(data);
+
+	// TODO: setup model, metric settings here. For some reason it's not working.
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[owner]/[project]/+layout.svelte:19]: More general question of how we want to do this. Some repos are "no TODO in prod", there should be a single source of truth (i.e., for us, linear issue). I also feel strongly about this because these TODO's become invisible pretty quickly and if everyone adds them they pollute code.

[Nice article on it](https://www.osedea.com/en/blog/should-you-put-todos-in-the-source-code).

Comment (cabreraalex) [reply frontend/src/routes/(app)/project/[owner]/[project]/+layout.svelte:19]: Sold!

Sparkier (approved):

PR #99, opened 2023-08-23
feat: dataset- and system-based Python SDK

cabreraalex:
# Description

New Python SDK for creating Zeno projects focused on batch upload of base datasets and system outputs.

PR #107, opened 2023-08-25
feat: update existing project upload example

cabreraalex:
# Description

Creates a Jupyter Notebook to upload existing Zeno caches.

# Blocked by

- #99


backend/zeno_backend/database/select.py
```diff
@@ -401,6 +401,24 @@ def project_stats(project: str) -> ProjectStats | None:
...
+    """
+    with Database() as db:
+        column_name = db.execute_return(
+            sql.SQL("SELECT column_id FROM {} WHERE type = 'DATA';").format(
+                sql.Identifier(f"{project_uuid}_column_map")
+            )
+        )
+        return str(column_name[0]) if column_name is not None else ""
+
+
```
Sparkier, line 421: We currently always name it the same, so we don't really need this. I think it can always be named `data_id`.

frontend/src/lib/components/instance-views/ListView.svelte
```diff
@@ -111,6 +111,9 @@
 						options={viewOptions}
 						entry={inst}
 						modelColumn={modelColumn?.id}
+						dataColumn={$columns.find((col) => col.columnType === ZenoColumnType.DATA)
+							? $columns.find((col) => col.columnType === ZenoColumnType.DATA)['name']
+							: ''}
```
Sparkier, line 116: Don't think we need this.

Sparkier (changes requested):

backend/zeno_backend/server.py
```diff
@@ -431,7 +440,7 @@ def login(name: str):
         else:
             return fetched_user
 
-    @api_app.post("/project", tags=["zeno"])
+    @api_app.post("/new-project", tags=["zeno"])
```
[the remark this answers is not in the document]
cabreraalex: I think standard for REST api is to not put actions, so "new" describes the endpoint better and "post" describes the action I think?

Sparkier (approved):

PR #109, opened 2023-08-25
fix: local fetch in docker

Sparkier:


Sparkier:
This needs some more testing. For me Local setup broke with this for some reason and Docker did not work either. Reverted for now will investigate later but not prio 1.

frontend/vite.config.ts
```diff
@@ -4,6 +4,12 @@ import { defineConfig } from 'vitest/config';
 export default defineConfig({
 	server: {
 		proxy: {
+			'/dockerzeno': {
+				target: 'http://backend:8000',
```
Sparkier, line 8: How does it know how to resolve `backend`?

cabreraalex (approved):

PR #114, opened 2023-08-25
feat: API key for Python SDK

cabreraalex:
# Description
Change the Python SDK to use an API key instead of Cognito authentication.

We want this eventually to be able to run Zeno workflows on servers/without a user, and it gets around some issues with cognito variables in the SDK code.


backend/zeno_backend/server.py
```diff
@@ -92,6 +82,7 @@ def get_server() -> FastAPI:
     api_app = FastAPI(
         title="Backend API", generate_unique_id_function=lambda route: route.name
     )
+    api_app.include_router(sdk.router)
```
Sparkier, line 85: Nice one, wanted to split up the two more as well. However, do we want to also differentiate the parent paths? Like we have `api` rn, do we want to add a separate called `cli` or something? Also, do we want these to have a tag as well. Like we have the 'zeno' tag do we want to add another one?

Sparkier (approved):
Have some comments on this but approving now.

1. capitalization: should address in a separate PR, created linear issue
2. api separation: if we change something here, can do in this PR

backend/zeno_backend/database/insert.py
```diff
@@ -14,20 +15,43 @@
...
     Creates a new entry in the projects table, creates a new table for the project's
     data, creates a table to map data columns to ids and types, and creates a table for
     all tags of the project.
 
     Args:
-        project_config (Project): the configuration with which to
+        project_config (Project): The configuration with which to
             initialize the new project.
-        user (User): the user who is setting up the project and becomes its admin.
+        owner_id (int): The id of the user who owns the project.
```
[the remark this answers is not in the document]
cabreraalex: Yeah don't feel strongly, lower is fine!

backend/zeno_backend/server.py
cabreraalex: hmm interesting point, maybe we make it /sdk or something?

cabreraalex: Will make a Linear issue :)

PR #123, opened 2023-08-28
fix: chart styling, table sorting

cabreraalex:
# Description

Update some chart home styling and fix sorting for table view.

### DIFF
--- a/frontend/src/lib/components/chart/ChartHomeBlock.svelte
+++ b/frontend/src/lib/components/chart/ChartHomeBlock.svelte
@@ -31,10 +31,9 @@
 	};
 </script>
 
-<div
+<button
 	class="flex flex-col items-center border-2 border-grey-lighter rounded-lg hover:bg-primary-ligther mt-2 cursor-pointer p-2 mr-2 max-w-[500px]"
-	on:click={() => goto(`${$page.url}/${chart.id}`)}
-	on:keypress={() => goto(`${$page.url}/${chart.id}`)}
+	on:click={() => goto(`${$page.url}/${chart.id}?edit=false`)}
 >
 	<div class="flex justify-between items-center w-full">
 		<div class="m-4 min-w-[24px]">
@@ -102,4 +101,4 @@
 			{/if}
 		</div>
 	</div>
-</div>
+</button>
--- a/frontend/src/lib/components/chart/ChartHomeBlock.svelte
+++ b/frontend/src/lib/components/chart/ChartHomeBlock.svelte
@@ -3,7 +3,7 @@
 		mdiBee,
 		mdiChartBar,
 		mdiChartLine,
-		mdiDotsVertical,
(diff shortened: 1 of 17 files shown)


frontend/src/lib/components/general/Header.svelte
```diff
@@ -43,18 +43,12 @@
...
-								/>
-							{/if}
-						{/await}
-					{/if}
+					<HeaderIcon
+						pageName={'chart'}
+						tooltipContent={'Create charts from your slices and metrics.'}
+						icon={mdiChartBoxOutline}
+						on:click={() => goto(`${getProjectRouteFromURL($page.url)}/chart`)}
+					/>
```
Sparkier, line 51: I only showed this for non-editors when there actually were charts. Otherwise they are brought to an empty page. Hence the numCharts check. Do we always want to show this? If we do, what do we show if there are no charts?

frontend/src/lib/components/chart/ChartHomeBlock.svelte
```diff
@@ -31,26 +31,28 @@
 	};
 </script>
 
-<button
-	class="flex flex-col items-center border-2 border-grey-lighter rounded-lg m-2 hover:bg-primary-light w-48 h-24"
+<div
+	class="flex flex-col items-center border-2 border-grey-lighter rounded-lg hover:bg-primary-ligther mt-2 cursor-pointer p-2 mr-2 max-w-[500px]"
 	on:click={() => goto(`${$page.url}/${chart.id}`)}
+	on:keypress={() => goto(`${$page.url}/${chart.id}`)}
```
Sparkier, line 37: Why make this a `div` when it acts as a button? I think this kinda destroys accessibility, no?

frontend/src/routes/(app)/project/[owner]/[project]/chart/[chartIndex=integer]/+page.svelte
```diff
@@ -12,7 +13,7 @@
 
 	export let data;
 
-	let isChartEdit = false;
+	let isChartEdit = $page.url.searchParams.get('edit') ? true : false;
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[owner]/[project]/chart/[chartIndex=integer]/+page.svelte:16]: If we do this via the URL, we should probably update the URL in the EditHeader as well instead of updating this variable probably. Then it works on reload as well.

Sparkier (changes requested):

frontend/src/lib/components/general/Header.svelte
cabreraalex: It was flickering every time on the sidebar when any interaction happened b/c of the fetch and it was bothering me aha. Maybe we can just put text in the tab if there are no charts.

Sparkier: Sounds good to me. Just wanted to bring this up since seeing a blank page might be weird.

Sparkier (approved):

PR #126, opened 2023-08-29
refactor: update sql calls to return list, fix route proxy

cabreraalex:
# Description
Change SQL calls to always return a list for consistency. 

Also update the endpoint routing to work with sveltekit in edge cases.


frontend/src/lib/util/util.ts
```diff
@@ -94,7 +95,10 @@ export function getEndpoint() {
 		return '/localzeno';
 	}
 	if (env.PUBLIC_BACKEND_ENDPOINT == 'http://zeno-backend:8000') {
```
Sparkier, line 97: Should we remove the localzeno stuff? Seems like this would be needed there as well otherwise.

Sparkier (approved):

PR #136, opened 2023-08-31
fix: text metadata visualizations, usability issues

cabreraalex:
# Description
Minor usability issues, show text filtering.

### DIFF
--- a/backend/zeno_backend/database/util.py
+++ b/backend/zeno_backend/database/util.py
@@ -24,7 +24,9 @@ def resolve_metadata_type(data_frame: pd.DataFrame, column: str) -> MetadataType
     elif pd.api.types.is_datetime64_any_dtype(dtype):
         return MetadataType.DATETIME
     elif pd.api.types.is_string_dtype(dtype):
-        return MetadataType.NOMINAL
+        if data_frame[column].nunique() < 20:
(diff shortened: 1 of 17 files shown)


backend/zeno_backend/database/util.py
```diff
@@ -24,7 +24,9 @@ def resolve_metadata_type(data_frame: pd.DataFrame, column: str) -> MetadataType
     elif pd.api.types.is_datetime64_any_dtype(dtype):
         return MetadataType.DATETIME
     elif pd.api.types.is_string_dtype(dtype):
-        return MetadataType.NOMINAL
+        if data_frame[column].nunique() < 20:
+            return MetadataType.NOMINAL
+        return MetadataType.OTHER
```
Sparkier, line 29: Right now we check this in the frontend. This would let us change this metric later but is more expensive. Not sure which one is better. This might be problematic if we ever want to change this number.

Sparkier (approved):

backend/zeno_backend/database/util.py
cabreraalex: Hmm good point...

cabreraalex: I'll leave a linear issue, not sure what the best solution is.

PR #151, opened 2023-09-05
fix: inconsistent server updates, don't run get after post, fix tags, 

cabreraalex:
# Description
Sveltekit was caching get requests for things like slices, causing it to not re-pull changes when tabs were changed. This is primarily fixed by calling `invalidateAll()`. This PR fixes that along with some other fixes/changes:

* Don't call get after post, just return ID and add locally for "post" calls
* Run initial fetch in parallel for performance
* Fix what is used to render data to separate from data_id for tags

Fixes:
[ZEN-92 : slices are not always updated on the frontend without refresh](https://linear.app/zenoml/issue/ZEN-92/slices-are-not-always-updated-on-the-frontend-without-refresh)
[ZEN-96 : metric dropdown is blank when navigating home and back to a project](https://linear.app/zenoml/issue/ZEN-96/metric-dropdown-is-blank-when-navigating-home-and-back-to-a-project)

### DIFF
--- a/backend/zeno_backend/server.py
+++ b/backend/zeno_backend/server.py
@@ -248,7 +248,7 @@ def get_server() -> FastAPI:
         # Prepend the DATA_URL to the data column if it exists
         project = select.project_from_uuid(project_uuid)
         if project and project.data_url:
-            filt_df["data_id"] = project.data_url + filt_df["data_id"]
+            filt_df["data"] = project.data_url + filt_df["data"]
 
         if req.diff_column_1 and req.diff_column_2:
             filt_df = generate_diff_cols(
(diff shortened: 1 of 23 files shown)


backend/zeno_backend/server.py
```diff
@@ -455,21 +455,65 @@ def login(name: str):
...
-    @api_app.post("/slice/{project}", tags=["zeno"], dependencies=[Depends(auth)])
+    @api_app.post(
+        "/slice/{project}",
+        response_model=int,
+        tags=["zeno"],
+        dependencies=[Depends(auth)],
+    )
     def add_slice(project: str, req: Slice):
-        insert.slice(project, req)
+        res = insert.slice(project, req)
```
Sparkier, line 480: Nit: call this id?

frontend/src/lib/components/chart/ChartHomeBlock.svelte
```diff
@@ -69,10 +69,17 @@
...
+										invalidateAll();
+										charts.update((c) => {
+											c.push({
+												id: res,
+												name: 'Copy of ' + chart.name,
+												type: chart.type,
+												parameters: chart.parameters
+											});
+											return c;
+										});
```
Sparkier, line 82: I fetched after updates to make sure we have the most recent result. Don't know if that's a big problem but this might be an issue with concurrent updates. Can probably merge for now though.

frontend/src/lib/components/metadata/cells/SliceFinderCell.svelte
```diff
@@ -29,28 +30,24 @@
...
 	function removeSlice() {
 		ZenoService.deleteSlice(slice).then(() => {
-			if ($project) {
-				ZenoService.getSlices($project.uuid).then((fetchedSlices) => {
-					slices.set(fetchedSlices);
-					created = false;
-				});
-			}
+			invalidateAll();
+			slices.update((s) => s.filter((sli) => sli.sliceName !== slice.sliceName));
```
Sparkier, line 49: Should not do by name imo. Can do by id, no?

frontend/src/routes/(app)/project/[owner]/[project]/+layout.server.ts
```diff
@@ -26,39 +34,26 @@ export async function load({ cookies, params, url }) {
...
-		models: models,
-		metrics: metrics,
-		folders: folders,
-		tags: tags,
+		slices: results[0] as Slice[],
+		columns: results[1] as ZenoColumn[],
+		models: results[2] as string[],
+		metrics: results[3] as Metric[],
+		folders: results[4] as Folder[],
+		tags: results[5] as Tag[],
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[owner]/[project]/+layout.server.ts:56]: Might be a bit problematic to find the reason if something fails but probably fine.

frontend/src/routes/(app)/project/[owner]/[project]/+layout.svelte
```diff
@@ -14,23 +16,27 @@
...
+	models.set(data.models);
+	if (data.models.length > 0) {
+		model.set(data.models[0]);
+	}
+	metrics.set(data.metrics);
+	if (data.metrics.length > 0) {
+		metric.set(data.metrics[0]);
+	}
+	folders.set(data.folders);
+	tags.set(data.tags);
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[owner]/[project]/+layout.svelte:32]: Not sure about this: previously was reactive to changes in data. But maybe reloads anyway if data changes?

Sparkier (changes requested):

frontend/src/routes/(app)/project/[owner]/[project]/+layout.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/project/[owner]/[project]/+layout.svelte:32]: Yeah if data changes it was reloaded.

frontend/src/lib/components/metadata/cells/SliceFinderCell.svelte
cabreraalex: hehe good catch, copilot messin me up.

Sparkier (approved):

PR #152, opened 2023-09-06
feat: encode state in URL

cabreraalex:
# Description
Encodes common state variables in the URL as a base64 JSON encoded string to enable sharing.

Depends on #151 
Fixes [ZEN-88 : encode state in URL](https://linear.app/zenoml/issue/ZEN-88/encode-state-in-url)


frontend/src/routes/(app)/project/[owner]/[project]/(metadata)/+layout.svelte
```diff
@@ -1,5 +1,8 @@
 <script lang="ts">
 	import MetadataPanel from '$lib/components/metadata/MetadataPanel.svelte';
+	import { decodeURLParameters } from '$lib/util/util';
+
+	decodeURLParameters();
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[owner]/[project]/(metadata)/+layout.svelte:5]: Should this happen on a higher level? What if we have params outside a project? We probably will in the future with reports.

frontend/src/routes/(app)/project/[owner]/[project]/(metadata)/[page=exploreOrCompare]/+page.svelte
```diff
@@ -1,7 +1,25 @@
...
 
 	export let data;
+
+	model.subscribe(() => setURLParameters());
+	metric.subscribe(() => setURLParameters());
+	comparisonModel.subscribe(() => setURLParameters());
+	comparisonColumn.subscribe(() => setURLParameters());
+	compareSort.subscribe(() => setURLParameters());
+	metricRange.subscribe(() => setURLParameters());
+	selections.subscribe(() => setURLParameters());
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[owner]/[project]/(metadata)/[page=exploreOrCompare]/+page.svelte:22]: Same, maybe do on a higher level?

Comment (Sparkier) [reply frontend/src/routes/(app)/project/[owner]/[project]/(metadata)/[page=exploreOrCompare]/+page.svelte:22]: And probably also in a +layout.svelte file?

Sparkier (changes requested):

frontend/src/routes/(app)/project/[owner]/[project]/(metadata)/+layout.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/project/[owner]/[project]/(metadata)/+layout.svelte:5]: hmm good point

Sparkier (approved):

PR #154, opened 2023-09-06
feat: refactor metrics folder and add BLEU score

cabreraalex:
# Description
Adds BLEU score to the backend and creates a new folder for metric calculation files.

Fixes [ZEN-100 : implement BLEU score](https://linear.app/zenoml/issue/ZEN-100/implement-bleu-score)

### DIFF
--- a/backend/zeno_backend/processing/metrics/map.py
+++ b/backend/zeno_backend/processing/metrics/map.py
@@ -1,0 +1,88 @@
+"""Map metric names to metric functions."""
+
+from psycopg import sql
+
+from zeno_backend.classes.base import GroupMetric
+from zeno_backend.classes.metric import Metric
+from zeno_backend.database.database import Database
+from zeno_backend.processing.metrics.bleu import bleu
+from zeno_backend.processing.metrics.f1 import f1, precision, recall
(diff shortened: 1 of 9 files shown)


backend/zeno_backend/processing/metrics/map.py
```diff
@@ -0,0 +1,88 @@
...
+        return mean(project, metric, model, sql_filter)
+    if metric.type == "bleu":
+        return bleu(project, metric, model, sql_filter)
+    if metric.type == "recall":
+        return recall(project, model, sql_filter)
+    if metric.type == "f1":
+        return f1(project, model, sql_filter)
+    if metric.type == "precision":
+        return precision(project, model, sql_filter)
+    if metric.type == "count":
```
Sparkier, line 85: Nit: should these be elif?

Sparkier (approved):

backend/zeno_backend/processing/metrics/map.py
cabreraalex: Just style wise? since they return it doesn't matter but maybe its better style?

Sparkier: Ah whops my bad, was late.

PR #155, opened 2023-09-06
fix: fix table filtering for output-dependent slices

cabreraalex:
# Description

For slices dependent on the output of a model in the compare tab, have to selectively filter for a selected model.

Fixes [ZEN-103 : Filtering doesn't work for slices with predicates dependent on the model in comparison tab](https://linear.app/zenoml/issue/ZEN-103/filtering-doesnt-work-for-slices-with-predicates-dependent-on-the)


cabreraalex:
This is a bit more complicated than I thought, worth rethinking or refactoring how we save slices.... the name of the column is tied to a model currently. Either have to store without a model reference or column name, or update them when doing calculations. A bit messy

cabreraalex:
Alright looks like @Sparkier had already sort of fixed this in the backend when creating filter_sql, just had to be updated for the table fetching with the current model. 

The only open issue is that filtering for a specific model in the comparison tab is very complicated and wasn't working (have to somehow specify in the `selection` store which model its for, would be a major refactor). I removed it for now, it only filters using Model A. Added a linear issue:

[ZEN-104 : comparison view only filters slices using first model](https://linear.app/zenoml/issue/ZEN-104/comparison-view-only-filters-slices-using-first-model)

Sparkier:
> This is a bit more complicated than I thought, worth rethinking or refactoring how we save slices.... the name of the column is tied to a model currently. Either have to store without a model reference or column name, or update them when doing calculations. A bit messy

Haha that's what I thought when first stumbling over this.

frontend/src/lib/util/charts.ts
```diff
@@ -12,12 +12,12 @@ import {
...
-import HeatMap from './chart-types/heatmap-chart/HeatMap.svelte';
-import LineChart from './chart-types/line-chart/LineChart.svelte';
-import RadarChart from './chart-types/radar-chart/RadarChart.svelte';
-import Table from './chart-types/table/Table.svelte';
+import BarChart from '../components/chart/chart-types/bar-chart/BarChart.svelte';
+import BeeswarmChart from '../components/chart/chart-types/beeswarm-chart/BeeswarmChart.svelte';
+import HeatMap from '../components/chart/chart-types/heatmap-chart/HeatMap.svelte';
+import LineChart from '../components/chart/chart-types/line-chart/LineChart.svelte';
+import RadarChart from '../components/chart/chart-types/radar-chart/RadarChart.svelte';
+import Table from '../components/chart/chart-types/table/Table.svelte';
```
Sparkier, line 20: This should probably be `$lib/components`

Sparkier (approved):

PR #157, opened 2023-09-07
feat: implement feature flag for the Zeno frontend

Sparkier:
# Description

Adds feature flags to Zeno's frontend. They need to be defined in `/src/lib/util/features.ts` and are stored in a cookie so that they are persisted for users.

image


Sparkier:
This is a very basic implementation of feature flags but I think we need **something** before we can add bigger new features that might not be ready for broad adoption yet.

cabreraalex:
What's the main goal of this? To have specific users use a feature? Or for us as developers to use it? Or for us to manage which features are in prod? 

If its one of the last two I think we might want another way of doing this where we can set features in some admin panel?

Sparkier:
What we probably ultimately want to be able to do is:
1. as devs activate/deactivate features on the fly (for this something in the UI would be good, maybe just in dev deploy)
2. activate/deactivate features for specific user groups

For 2. I think we need bigger changes and a broader discussion on how to do this. This is just a quick implementation to get feature flags. It should not prevent us from moving to something else in the future. I think the most important thing is to have *something* for now. Changing the implementation will be pretty easy.

cabreraalex:
SGTM!

cabreraalex (approved):

PR #158, opened 2023-09-08
fix: make string filtering work

cabreraalex:
# Description
Simplify to regex or case sensitivity.

### DIFF
--- a/frontend/src/lib/components/popups/FilterEntry.svelte
+++ b/frontend/src/lib/components/popups/FilterEntry.svelte
@@ -75,7 +75,7 @@
 	</div>
 	<div class="mr-2.5">
 		{#if predicate.column}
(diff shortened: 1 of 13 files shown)


frontend/src/lib/components/popups/FilterEntry.svelte
```diff
@@ -84,17 +84,26 @@
...
 					renderer={renderOptions}
 				/>
+			{:else}
+				<Svelecte
+					on:change={operationChange}
+					value={inverseOperationMap[predicate.operation]}
+					valueField="label"
+					placeholder={'Operation'}
+					searchable={false}
+					options={['==', '!=', 'LIKE', 'ILIKE', 'REGEX']}
```
Sparkier, line 105: Can we get this from the operation enum?

Sparkier (approved):

Sparkier:
Ah I was thinking automatically all (or all we need).

PR #159, opened 2023-09-08
refactor: fix histogram processing

cabreraalex:
# Description
Simplify histogram generation, only one function to generate buckets and another to calculate counts & metrics.

### DIFF
--- a/backend/zeno_backend/database/util.py
+++ b/backend/zeno_backend/database/util.py
@@ -18,6 +18,8 @@ def resolve_metadata_type(data_frame: pd.DataFrame, column: str) -> MetadataType
     """
     dtype = data_frame[column].dtype
     if pd.api.types.is_any_real_numeric_dtype(dtype):
+        if data_frame[column].nunique() < 20:
+            return MetadataType.NOMINAL
         return MetadataType.CONTINUOUS
     elif pd.api.types.is_bool_dtype(dtype):
         return MetadataType.BOOLEAN
--- a/frontend/src/lib/components/general/Header.svelte
+++ b/frontend/src/lib/components/general/Header.svelte
@@ -70,7 +70,7 @@
 			{/if}
 		</div>
 		<div class="flex flex-col items-center justify-center mb-3">
-			{#if currentTab?.includes('project') && $project && $project.ownerName === user?.name}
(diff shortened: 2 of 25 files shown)


frontend/src/lib/components/metadata/HistogramsHeader.svelte
```diff
@@ -8,7 +8,7 @@
 </script>
 
 <div
-	class="flex items-center justify-between pt-2.5 sticky top-24 border-b border-grey-lighter bg-yellowish-light z-10 pb-4"
+	class="flex items-center justify-between pt-2.5 top-24 border-b border-grey-lighter bg-yellowish-light z-10 pb-4"
```
Sparkier, line 11: Not sure but if not sticky can we remove the top property as well?

frontend/src/lib/components/general/Header.svelte
```diff
@@ -70,7 +70,7 @@
 			{/if}
 		</div>
 		<div class="flex flex-col items-center justify-center mb-3">
-			{#if currentTab?.includes('project') && $project && $project.ownerName === user?.name}
+			{#if (currentTab?.includes('explore') || currentTab?.includes('compare')) && $project?.ownerName === user?.name}
```
Sparkier, line 73: Don't we want to show this for chart? Seems like something people might want to access after chart creation.

Sparkier (approved):

backend/zeno_backend/database/util.py
```diff
@@ -18,6 +18,8 @@ def resolve_metadata_type(data_frame: pd.DataFrame, column: str) -> MetadataType
     """
     dtype = data_frame[column].dtype
     if pd.api.types.is_any_real_numeric_dtype(dtype):
+        if data_frame[column].nunique() < 20:
```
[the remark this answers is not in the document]
cabreraalex: ugh I know it's terrible let me think a bit more about it. There are so many weird edge cases...

frontend/src/lib/components/general/Header.svelte
cabreraalex: Good catch, it wasn't working at all before when checking for project.

backend/zeno_backend/database/util.py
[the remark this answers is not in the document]
cabreraalex: idk how to do this well, going to make a linear issue.

PR #166, opened 2023-09-11
fix: speed up routes

cabreraalex:
# Description
Various changes to improve performance.
* Consolidate get routes for state and charts
* Use await and promises for table and results for consistency


backend/zeno_backend/server.py
```diff
@@ -175,14 +175,19 @@ def get_slices(project: str, request: Request):
...
     @api_app.get(
-        "/charts/{project}",
+        "/charts/{owner}/{project}",
         response_model=list[Chart],
         tags=["zeno"],
     )
-    def get_charts(project: str, request: Request):
-        if not util.access_valid(project, request):
+    def get_charts(owner_name: str, project_name: str, request: Request):
+        project_uuid = select.project_uuid(owner_name, project_name)
```
Sparkier, line 183: Can't we do this from the project uuid? Should have this already when fetching charts, no?

backend/zeno_backend/server.py
```diff
@@ -260,15 +265,30 @@ def get_filtered_table(req: TableRequest, project_uuid: str, request: Request):
             )
         return filt_df.to_json(orient="records")
 
-    @api_app.post(
-        "/chart-data/{project}",
+    @api_app.get(
+        "/chart/{owner}/{project}/{chart_uuid}",
```
Sparkier, line 269: Nit: Charts don't have a uuid but an id, no?

backend/zeno_backend/server.py
```diff
@@ -260,15 +265,30 @@ def get_filtered_table(req: TableRequest, project_uuid: str, request: Request):
...
+        response_model=ChartResponse,
         tags=["zeno"],
-        response_model=str,
     )
-    def get_chart_data(chart: Chart, project: str, request: Request):
-        if not util.access_valid(project, request):
+    def get_chart(owner_name: str, project_name: str, chart_id: int, request: Request):
+        project_uuid = select.project_uuid(owner_name, project_name)
```
Sparkier, line 274: Same as above.

cabreraalex: We don't necessarily have it. We could wait for the parent layout to resolve, but then it causes a waterfall and is slow.

PR #167, opened 2023-09-11
fix: update when histogram and table calls are made

cabreraalex:
# Description
This PR uses derived stores and makes some updates to make fewer redundant requests to the backend and speed up processing across the app.

### DIFF
--- a/frontend/src/routes/(app)/project/[owner]/[project]/+layout.server.ts
+++ b/frontend/src/routes/(app)/project/[owner]/[project]/+layout.server.ts
@@ -14,8 +14,6 @@ export async function load({ cookies, params, url, depends }) {
 
 	OpenAPI.BASE = getEndpoint() + '/api';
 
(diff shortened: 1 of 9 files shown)


frontend/src/routes/(app)/project/[owner]/[project]/+layout.server.ts
```diff
@@ -27,8 +25,6 @@ export async function load({ cookies, params, url, depends }) {
 		OpenAPI.HEADERS = {
 			Authorization: 'Bearer ' + cognitoUser.accessToken
 		};
-	} else if (!projectPublic) {
-		throw redirect(303, `/login?redirectTo=${url.pathname}`);
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[owner]/[project]/+layout.server.ts:31]: Are we doing this somewhere else now? Maybe I missed it.

Sparkier (approved):

PR #172, opened 2023-09-14
fix: fix LRU cache

cabreraalex:
# Description
LRU cache was not implemented correctly. This fixes it by making all parameters to metric function hasheable.

### DIFF
--- a/backend/zeno_backend/classes/filter.py
+++ b/backend/zeno_backend/classes/filter.py
@@ -1,7 +1,9 @@
 """Classes required for specifying data filters."""
 import json
 from enum import Enum
-from typing import LiteralString
+from typing import TYPE_CHECKING, LiteralString
+
+from pydantic import ConfigDict
 
 from zeno_backend.classes.base import CamelModel, ZenoColumn
 
@@ -55,6 +57,8 @@ class Join(str, Enum):
 class FilterPredicate(CamelModel):
     """Predicates to specify a filter operation on a column of the data table."""
 
+    model_config = ConfigDict(frozen=True)
+
     column: ZenoColumn
     operation: Operation
     value: str | float | int | bool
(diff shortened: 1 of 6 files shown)


backend/zeno_backend/classes/filter.py
```diff
@@ -64,9 +68,16 @@ class FilterPredicate(CamelModel):
 class FilterPredicateGroup(CamelModel):
     """Group of filter predicates that might be joined by a Join operator."""
 
-    predicates: list["FilterPredicateGroup | FilterPredicate"]
+    model_config = ConfigDict(frozen=True)
+
+    predicates: tuple["FilterPredicateGroup | FilterPredicate"]
```
Sparkier, line 73: Why tuple?

backend/zeno_backend/processing/metrics/map.py
```diff
@@ -55,23 +57,32 @@ def count(
...
     Returns:
         GroupMetric: the metric result calculated on the data as specified.
     """
+    sql_filter = table_filter(
+        project,
+        model,
+        filter_predicates,
+        list(data_ids) if data_ids is not None else None,
+    )
+    print(sql_filter)
```
Sparkier, line 85: Remove.

frontend/src/lib/zenoapi/models/FilterPredicateGroup.ts
```diff
@@ -3,13 +3,12 @@
...
 
-import type { FilterPredicate } from './FilterPredicate';
 import type { Join } from './Join';
 
 /**
  * Group of filter predicates that might be joined by a Join operator.
  */
 export type FilterPredicateGroup = {
-	predicates: Array<FilterPredicateGroup | FilterPredicate>;
+	predicates: any[];
```
Sparkier, line 12: Is this really necessary? Would be much better if this was not any. We're using it in the front end as well...

Sparkier (changes requested):

PR #173, opened 2023-09-14
feat: move cache to chart data

cabreraalex:
# Description
Move LRU cache to cache table data, which is simpler than caching metric calls.

### DIFF
--- a/frontend/src/lib/components/chart/chart-types/radar-chart/RadarChart.svelte
+++ b/frontend/src/lib/components/chart/chart-types/radar-chart/RadarChart.svelte
@@ -13,8 +13,8 @@
 			size: number;
 		}>;
 	};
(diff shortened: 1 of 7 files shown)


frontend/src/lib/components/chart/chart-types/radar-chart/RadarChart.svelte
```diff
@@ -13,8 +13,8 @@
 			size: number;
 		}>;
 	};
-	export let width = 1000;
-	export let height = 400;
+	export let width = 700;
+	export let height = 600;
```
Sparkier, line 17: Nit: should not do changes like these in PRs about something completely different probably.

Sparkier (approved):

PR #174, opened 2023-09-15
feat: db pool for histogram calc

cabreraalex:
# Description
Actually use asyncio for DB calculations using an async DB pool. Speeds up by a factor of features being calculated, e.g. 5x for translation.


cabreraalex:
okay this is actually ready now. Very odd asyncio event loop issues in prod I was dealing with.

backend/Dockerfile.prod
```diff
@@ -10,4 +10,4 @@ RUN poetry install
 
 EXPOSE 80
 
-CMD ["poetry", "run", "backend"]
\ No newline at end of file
+CMD exec poetry run uvicorn zeno_backend.server:get_server --reload --host 0.0.0.0 --port 80
```
cabreraalex, line 13: Running it as a module didn't create an event loop in asyncio oddly. Also matches the dev setup.

backend/zeno_backend/database/database.py
```diff
@@ -6,6 +6,49 @@
...
+        db: dict[str, Any] = {}
+        db["host"] = os.environ["DB_HOST"]
+        db["port"] = os.environ["DB_PORT"]
+        db["dbname"] = os.environ["DB_NAME"]
+        db["user"] = os.environ["DB_USER"]
+        db["password"] = os.environ["DB_PASSWORD"]
+        return db
+
+
+db_pool = AsyncConnectionPool(" ".join([f"{k}={v}" for k, v in config().items()]))
```
cabreraalex, line 51: Eventually probably want to switch most queries to this, but it's okay to have both for now IMO

Sparkier (approved):

PR #176, opened 2023-09-15
feat: re-add frontend amplitude and simplify backend

cabreraalex:
# Description
Our backend calls aren't consistent or ID-ed enough to do real tracking. I also am not too concerned about GDPR with amplitude at our scale, so think the frontend tracking is okay. Eventually we should have some cookie banner.


backend/zeno_backend/server.py
```diff
@@ -344,24 +344,6 @@ def get_project_state(
                 project.editor = True
         return select.project_state(project_uuid, project)
 
-    @api_app.post("/project/{owner}/{project}", response_model=Project, tags=["zeno"])
-    def get_project(owner_name: str, project_name: str, request: Request):
```
cabreraalex, line 348: no longer called with the big project_stats call

frontend/src/routes/(user)/signup/+page.svelte
```diff
@@ -48,7 +49,14 @@
 			label="Repeat password"
 			class="w-56 mb-3"
 		/>
-		<Button type="submit" variant="raised" class="mt-5">Sign Up</Button>
+		<Button
+			type="submit"
+			variant="raised"
+			class="mt-5"
+			on:click={() => amplitude.track('User Registered')}
```
Sparkier, line 56: Shouldn't we track this only if registration was successful?

frontend/src/routes/+layout.svelte
```diff
@@ -1,10 +1,17 @@
...
 	import { featureFlags } from '$lib/stores';
 	import { zenoFeatureFlags } from '$lib/util/features';
+	import * as amplitude from '@amplitude/analytics-browser';
 	import '../app.css';
 
 	export let data;
 	featureFlags.set({ ...zenoFeatureFlags, ...data.features });
+
+	if (browser)
+		amplitude.init('9e6755badfe6f970b509d3325a6a6678', {
```
Sparkier, line 12: Should this be in here? I'd guess we might want different ones depending on dev/prod and also custom installations. Further, this should probably not be exposed, should it?

backend/zeno_backend/server.py
```diff
@@ -5,7 +5,7 @@
 
 import pandas as pd
 import uvicorn
-from amplitude import Amplitude, BaseEvent
+from amplitude import Amplitude
```
Sparkier, line 8: Can't this be removed completely?

Sparkier (changes requested):

frontend/src/routes/+layout.svelte
cabreraalex: re: dev vs. prod yeah maybe we want to read from ENV but changing ENV variables is a pain in prod aha.

It's okay to be exposed, it's usually just in the raw HTML.

frontend/src/routes/(user)/signup/+page.svelte
cabreraalex: Hmm probably, but the amplitude events don't work well in the backend js. Maybe since we track login events amplitude can infer registrations and we can remove this.

frontend/src/routes/+layout.svelte
cabreraalex: made it an env variable

Sparkier (approved):

PR #177, opened 2023-09-16
fix: move references to amplitude key to after .env loading

Sparkier:
Previously, the environment variable containing the amplitude API key was referenced before loading the .env file. This has been changed so that the key does not have to be set outside the .env file.

### DIFF
--- a/backend/zeno_backend/routers/sdk.py
+++ b/backend/zeno_backend/routers/sdk.py
@@ -20,8 +20,6 @@ from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 from zeno_backend.classes.project import Project
 from zeno_backend.database import insert, select
 
-amplitude_client = Amplitude(os.environ["AMPLITUDE_API_KEY"])
-
 
 class APIKeyBearer(HTTPBearer):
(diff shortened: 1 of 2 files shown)


Sparkier:
I also think that we should make this key entirely optional since people might want to deploy without using amplitude.

[Created a linear issue.](https://linear.app/zenoml/issue/ZEN-128/make-amplitude-api-key-optional)

cabreraalex:
idk if we want to initialize amplitude in each server call? probably only want to start it when the server starts

Sparkier:
> idk if we want to initialize amplitude in each server call? probably only want to start it when the server starts

I agree, but if we do it like now the server crashes if the env variable is not set since we try to access it before loading the `.env` file. So this is more of a hot fix for that.

Sparkier:
What we could do instead is check for that env variable before initializing amplitude and only log things when it's set. This would also not log events when developing locally and might be the preferable option.

Sparkier:
I changed the implementation to use a singleton object. This also checks if there is a tracking key at all.

cabreraalex (approved):

PR #190, opened 2023-09-18
feat: fix up charts in reports

cabreraalex:
# Description
Make all charts work reasonable in reports, mostly by changing store references to DB calls.


frontend/src/lib/components/chart/chart-types/table/SliceDetailsContainer.svelte
```diff
@@ -14,14 +10,6 @@
...
 	<button
 		class="text-primary"
-		on:click={() => {
-			selections.update((sel) => ({
-				slices: sli !== undefined && sli.sliceName !== 'All Instances' ? [sli.id] : [],
-				metadata: sel.metadata,
-				tags: sel.tags
-			}));
-			goto(`${getProjectRouteFromURL($page.url)}/explore`);
-		}}
```
Sparkier, line 24: If we remove this, should we not make it a button anymore maybe?

frontend/src/lib/components/report/ChartElement.svelte
```diff
@@ -32,7 +32,7 @@
 			{chart}
 			data={JSON.parse(data)}
 			width={650}
-			height={300}
+			height={chart.type == ChartType.RADAR ? 600 : 300}
```
Sparkier, line 35: Kinda weird to have it hard-coded like this but probably fine for now...

Sparkier (approved):

frontend/src/lib/components/report/ChartElement.svelte
cabreraalex: Yeah sizing across the board for charts is wonky we should standardize it somehow but it's surprisingly tricky.

PR #193, opened 2023-09-19
fix: remove stray layout file

cabreraalex:
# Description
Breaking build in prod

### DIFF
--- a/frontend/Dockerfile.prod
+++ b/frontend/Dockerfile.prod
@@ -13,6 +13,8 @@ RUN npm install
 # Copy the rest of the files
 COPY . .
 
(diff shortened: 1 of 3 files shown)


frontend/Dockerfile.prod
```diff
@@ -13,6 +13,8 @@ RUN npm install
 # Copy the rest of the files
 COPY . .
 
+ENV PUBLIC_AMPLITUDE_API_KEY=""
+
```
Sparkier, line 17: Do we always nee this? Long term would hope it works without this env var.

Sparkier (approved):

PR #198, opened 2023-09-20
fix: address multiple authentication issues

Sparkier:
# Description
- correctly checks whether an authenticated user can access a project
- waits for cookie reset when logging out
- checks and resets the authentication token wherever the user cookie is
  needed
- correctly checks whether an authenticated user can access a report


Sparkier:
I think this fixes most of our authentication issues. However, it's pretty hard to test and reproduce them so not quite sure...

cabreraalex (approved):

PR #216, opened 2023-09-25
fix: remove /ping route from logging

Sparkier:
Fixes ZEN-146


neubig (approved):
Looks great, thank you!

PR #218, opened 2023-09-25
feat: check if view is valid in client

cabreraalex:
# Description
Fix ZEN-157


Sparkier (approved):
Kinda weird to have to update in two places but don't really see a better way either.

PR #224, opened 2023-09-26
feat: order slices so that they are fetched according to their id order

Sparkier:
fix ZEN-139


Sparkier:
@cabreraalex this fetches slices in the correct order but somehow the VegaLite component does not re-render even though the data changes.

cabreraalex (approved):

PR #226, opened 2023-09-26
fix: cast data column to string always

Sparkier:
# Description

When the data column did contain non-string data (e.g. arrays), the SQL alchemy insert was failing. Since the data column can be narrowed by the instance view in all cases, we can just store strings and extract information in the instance view.

fix ZEN-153
fix ZEN-147


neubig (approved):
LGTM, thanks!

PR #229, opened 2023-09-27
feat: add back slice sorting with typecast

Sparkier:
fix ZEN-167

### DIFF
--- a/backend/zeno_backend/database/select.py
+++ b/backend/zeno_backend/database/select.py
@@ -985,10 +985,12 @@ def slices(project: str, ids: list[int] | None = None) -> list[Slice]:
         )
     else:
         slice_results = db.connect_execute_return(
-            "SELECT id, name, folder_id, filter "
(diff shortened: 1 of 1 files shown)


Sparkier:
@cabreraalex We should test this on dev before deploying to prod.

Sparkier:
Yes but with a type cast. Maybe this works. Need to test.

Sparkier:
I tried a bunch of stuff to make encodings change on chart type change, but this is a bit problematic. Hence, I had to implement a hack to force the encoding component to re-render. The problem is the following:

Svelecte can only be used with dndzone when using `bind:value`, not with `on:change`, as `on:change` does not get triggered on reordering on elements with dndzone. However, we can't bind:value always because otherwise we have an infinite loop (value change leads to a dispatch so we can change the chart which leads to a value change).

cabreraalex (approved):

PR #236, opened 2023-09-27
fix: difference column for non-numeric

cabreraalex:
# Description
fix comparison for non-numeric

Fix ZEN-172
Fix ZEN-181


neubig (approved):
Looks great, thanks!

PR #241, opened 2023-09-28
feat: consolidate columns in comparison and fix

cabreraalex:
# Description
Makes the feature columns in the same column as instances.

Also fixes ZEN-184


cabreraalex:
Screenshot 2023-09-28 at 20 09 17

Sparkier (approved):

PR #247, opened 2023-09-29
fix: small rendering fixes openai-chat

cabreraalex:
# Description
small formatting issues in markdown chat.

### DIFF
--- a/frontend/src/lib/components/instance-views/views/views/OpenAIChatMarkdown.svelte
+++ b/frontend/src/lib/components/instance-views/views/views/OpenAIChatMarkdown.svelte
@@ -36,7 +36,7 @@
 	<CircularProgress style="height: 32px; width: 32px; margin-right:20px" indeterminate />
 {:then data}
 	{@const entries = showall ? data : data.slice(-4)}
-	<div class="flex flex-col border border-grey-light rounded p-2.5 m-1 w-[32rem]">
+	<div class="flex flex-col border border-grey-light rounded p-2.5 m-1">
 		{#if !showall}
 			<button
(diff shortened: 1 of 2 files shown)


frontend/src/lib/components/instance-views/views/views/OpenAIChatMarkdown.svelte
```diff
@@ -36,7 +36,7 @@
 	<CircularProgress style="height: 32px; width: 32px; margin-right:20px" indeterminate />
 {:then data}
 	{@const entries = showall ? data : data.slice(-4)}
-	<div class="flex flex-col border border-grey-light rounded p-2.5 m-1 w-[32rem]">
+	<div class="flex flex-col border border-grey-light rounded p-2.5 m-1">
```
Sparkier, line 39: Doesn't it grow super wide now?

Sparkier (approved):

frontend/src/lib/components/instance-views/views/views/OpenAIChatMarkdown.svelte
cabreraalex: Yeah I think it's good for this one though since the markdown could be huge? maybe we set a bigger max?

Sparkier: Sure can do. Just was afraid that now we'd have only one super wide instance on screen 😂

cabreraalex: set a bigger max - still covers most of screen but in a good way ;)

PR #260, opened 2023-10-02
feat: add route for determining the minimum version number of the client

Sparkier:
towards ZEN-195


cabreraalex (approved):

cabreraalex:
This is gold for the change we about to push

PR #278, opened 2023-10-06
feat: add instance view for RAG

Sparkier:
fix ZEN-202

### DIFF
--- a/backend/zeno_backend/routers/sdk.py
+++ b/backend/zeno_backend/routers/sdk.py
@@ -31,6 +31,7 @@ VIEWS = [
     "openai-chat",
     "space-separated-values",
     "text-classification",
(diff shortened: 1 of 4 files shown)


cabreraalex:
could you add a screenshot or example data to test this? And could you pair it with a PR to https://github.com/zeno-ml/zenoml.com/pulls to add it to the website with an example of the data format?

Sparkier:
Yep, 100% should have done that.

image

https://github.com/zeno-ml/zenoml.com/pull/20

Sparkier:
An answer and a label can also be displayed. Those are plain text similar to the input section. Documents and the answer are both expected to be in the output column.

cabreraalex (approved):

PR #286, opened 2023-10-09
feat: add option to add slices for all values of a nominal column

Sparkier:
# Description

The slice popup changes if a user selects only a nominal column without entering anything else as follows:
image

After creation, slices will look as follows:
image

fix ZEN-214

### DIFF
--- a/backend/zeno_backend/database/delete.py
+++ b/backend/zeno_backend/database/delete.py
@@ -61,19 +61,29 @@ def report(report_id: int):
     )
 
 
-def folder(folder: Folder):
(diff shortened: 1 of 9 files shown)


cabreraalex:
should we add a three dot menu next to each histogram for this? And do we want to do this for the continuous histograms?

Sparkier:
Something I thought about as well but was not sure how to do. Right now it's in the slice popup but I think I agree it might be better next to histograms. Don't know about continuous either. It's not really meaningful for continuous histograms is it?

cabreraalex (approved):

PR #289, opened 2023-10-09
feat: add instance view to reports

cabreraalex:
# Description
Adds a new element type to reports to show raw instances.

Fix ZEN-123

Screenshot 2023-10-09 at 16 48 08


backend/zeno_backend/classes/report.py
```diff
@@ -10,10 +10,12 @@ class ReportElementType(Enum):
     Attributes:
         CHART: chart element for a report.
         TEXT: text element for a report.
+        INSTANCES: list of instances element for a report.
```
cabreraalex, line 13: Could be "SLICE" but we might support tags

backend/zeno_backend/database/select.py
```diff
@@ -1336,38 +1527,6 @@ def table_data(project: str, filter_sql: sql.Composed | None) -> SQLTable:
         )
 
 
-def column_id_from_name_and_model(
```
cabreraalex, line 1339: Moved b/c of circular import

Sparkier (changes requested):
I would rename your `Instances` to `Slice` and add another type once we want to support tags. All of this seems pretty specific to slices and I would assume that we need to write something similar but substantially different for tags anyway. This would also make it easier in some places to infer if we're talking about slices or tags rn for report instances.

Sparkier (approved):

PR #290, opened 2023-10-10
feat: collapse openai chat

cabreraalex:
# Description
Fix ZEN-223


Sparkier:
Can you add this to the RAG view as well? Don't know if we are using the expand thing somewhere else.

Sparkier (approved):

PR #292, opened 2023-10-10
fix: sharing reports

cabreraalex:
# Description
sharing reports broke homescreen due to indexing issues


neubig (approved):

neubig:
This is simple so I just reviewed it.

PR #293, opened 2023-10-11
feat: add addslice button to selection predicates

Sparkier:
# Description

image

fix ZEN-226


cabreraalex:
should we make it purple?

Sparkier:
You mean making it `raised`?

cabreraalex:
or unelevated

cabreraalex:
also if we put it here should we remove it from the other place? or okay in both?

Sparkier:
Don't know, looks a bit weird to me:
image

Sparkier:
> from

Was thinking about this as well. Don't know what one would expect. It does not really hurt to have it in both places right?

cabreraalex:
okay maybe white bg and keep both

cabreraalex (approved):

PR #297, opened 2023-10-11
feat: update report editing workflow

cabreraalex:
# Description
Update report editing to match designs.

Fix ZEN-235

https://github.com/zeno-ml/zeno-hub/assets/4563691/8f97867a-3d69-41d1-9aa7-28cb7df3e2b7


frontend/src/lib/components/popups/Confirm.svelte
```diff
@@ -14,21 +14,25 @@
...
-				<Button style="margin-right: 10px" variant="outlined" on:click={() => dispatch('confirm')}>
-					{confirmText}
-				</Button>
+			<div class="flex m-auto justify-center">
 				<Button style="margin-right: 10px" variant="outlined" on:click={() => dispatch('cancel')}>
 					{cancelText}
 				</Button>
+				<Button
+					style="margin-right: 10px"
+					variant="unelevated"
```
Sparkier, line 31: I would undo this because it's the only place we use this rn and the button is a different size which makes it look super weird.
Can fix this here if we want this style but would prefer a separate PR and using it as the confirm button in all dialogs.

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
```diff
@@ -94,27 +83,31 @@
...
 	<div
 		class="flex flex-col max-w-4xl m-auto bg-background px-10 pb-20 md:mt-6 md:mb-6 sm:mt-0 sm:mb-0 rounded shadow"
 	>
 		<div class="flex items-center mt-12 justify-between">
-			<h1
-				class="text-5xl mr-6 text-grey-darkest"
-				contenteditable={isEdit ? true : false}
-				on:blur={(e) => updateReportName(e)}
-			>
+			<h1 class="text-5xl mr-6 text-grey-darkest">
```
Comment (Sparkier) [inline frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:103]: How can one update the name of a report now?

frontend/src/lib/components/popups/Confirm.svelte
```diff
@@ -14,21 +14,25 @@
 </script>
 
 <div
-	class="absolute inset-0 z-20 flex justify-center items-baseline p-12 bg-grey bg-opacity-60"
+	class="fixed inset-0 z-20 flex justify-center items-baseline p-12 bg-grey bg-opacity-60"
```
Sparkier, line 17: Good catch, I think this is wrong for all other popups as well.

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
```diff
@@ -127,31 +120,31 @@
...
+					bind:showConfirmDelete
+					{addElement}
+					{chartOptions}
+					{sliceOptions}
+					report={data.report}
+				/>
 			{/each}
-			{#if isEdit}
-				<Button variant="raised" on:click={() => addElement(elements.length)}>Add Element</Button>
-			{/if}
```
Comment (Sparkier) [inline frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:154]: I think there is no way for me to add an element if there is no element in the report already now.

Sparkier (changes requested):

frontend/src/lib/components/popups/Confirm.svelte
```diff
@@ -14,21 +14,25 @@
...
+	<Paper class="p-6 flex flex-col" elevation={7}>
 		<Content>
-			<span>{message}</span>
+			<div class="mb-4">{message}</div>
 			<slot />
-			<div class="flex flex-row-reverse">
-				<Button style="margin-right: 10px" variant="outlined" on:click={() => dispatch('confirm')}>
-					{confirmText}
-				</Button>
+			<div class="flex m-auto justify-center">
```
Sparkier, line 25: We place the buttons in all popups on the right. Would keep that behavior since it's also what most apps do and what's best [in terms of usability](https://uxmovement.com/buttons/why-ok-buttons-in-dialog-boxes-work-best-on-the-right/).

frontend/src/lib/components/popups/Confirm.svelte
cabreraalex: SG, we should change the color though it's hard to tell at a glance which is which.

cabreraalex: [ZEN-241 : update confirm button style](https://linear.app/zenoml/issue/ZEN-241/update-confirm-button-style)

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:103]: In the settings!

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:154]: LOL good catch

Sparkier (approved):

PR #298, opened 2023-10-12
fix: massive fix to remove vulnerability leaking auth data

cabreraalex:
# Description
The simple `OpenAPI.HEADER = ` calls were saved across backend calls, leading to authentication sticking to the most recent user. 

This is a massive refactor that changes the OpenAPI to be an object instead of a global, which is set by the backend server routes or used as a context in the browser.

Fixes many of the issues we were having with authentication.

Fixes ZEN-183
Fixes ZEN-212


backend/zeno_backend/processing/filtering.py
```diff
@@ -117,8 +117,17 @@ def table_filter(
...
-            sql.SQL(",").join(map(sql.Literal, data_ids))
+        db = Database()
+        id_column = db.connect_execute_return(
+            sql.SQL("SELECT column_id FROM {} WHERE type = 'ID';").format(
+                sql.Identifier(f"{project}_column_map")
+            ),
+        )
+        if len(id_column) == 0:
+            return None
+        datapoint_filter = sql.SQL("{} IN ({})").format(
```
cabreraalex, line 128: This fixes a regression for tags

frontend/OpenAPI_back.txt
cabreraalex, line 1: no longer need this hack

frontend/package.json
```diff
@@ -13,8 +13,7 @@
 		"lint": "prettier --plugin-search-dir . --check . && eslint .",
 		"format": "prettier --plugin-search-dir . --write .",
 		"smui-theme": "smui-theme compile static/smui.css -i src/theme",
-		"generate-api": "npx openapi-typescript-codegen --input http://127.0.0.1:8000/api/openapi.json --output ./src/lib/zenoapi --client fetch; npx prettier -w ./src/lib/zenoapi && npm run copy-openapi",
-		"copy-openapi": "shx cp ./OpenAPI_back.txt ./src/lib/zenoapi/core/OpenAPI.ts"
+		"generate-api": "npx openapi-typescript-codegen --input http://127.0.0.1:8000/api/openapi.json --output ./src/lib/zenoapi --name ZenoClient; npx prettier -w ./src/lib/zenoapi "
```
cabreraalex, line 16: then `--name` command changes it to be an object instead of a global

frontend/src/routes/api/features/+server.ts
```diff
@@ -5,7 +5,6 @@ export async function POST({ request, cookies }) {
 	const data = await request.json();
 	cookies.set('featureFlags', JSON.stringify(data), {
 		path: '/',
-		httpOnly: true,
```
cabreraalex, line 8: this is the default

frontend/src/routes/(user)/login/+page.svelte
```diff
@@ -13,6 +20,7 @@
 	use:enhance
 >
 	<div class="flex flex-col items-center">
+		<input type="hidden" name="redirect" value={redirect} />
```
cabreraalex, line 23: the form was a different URL so this wasn't sent.

Sparkier (approved):

PR #299, opened 2023-10-12
feat: add delete confirmation popup when deleting something

Sparkier:
fix ZEN-230


cabreraalex:
do we want this for slices, tags, and report elements? Can be a bit annoying if you are deleting a bunch in a row. but maybe worth it?

Sparkier:
Was thinking the same and don't have a definitive answer. One hand can be annoying, other hand if you delete something accidentally from a report by clicking the X this can be pretty devastating.

Sparkier:
Same for carefully curated slices and tags.

cabreraalex:
Okay, LGTM. We can add a "don't ask again" button eventually lol

cabreraalex (approved):

PR #301, opened 2023-10-12
feat: redirect to user home if no other direction is set

Sparkier:


cabreraalex (approved):

Sparkier:
Hmm I was always redirected to discover before, which annoyed me.

PR #303, opened 2023-10-12
fix: prevent illegal project or report names by removing certain char…

Sparkier:
…acters and encoding the URI

fix ZEN-233

Depends on:
https://github.com/zeno-ml/zeno-client/pull/31
and
https://github.com/zeno-ml/zeno-hub/pull/302

### DIFF
--- a/backend/zeno_backend/database/select.py
+++ b/backend/zeno_backend/database/select.py
@@ -529,7 +529,7 @@ def report(
             [report_name, owner_id],
         )
         if len(report_result) == 0:
-            raise Exception("Project does not exist.")
+            raise Exception("Report does not exist.")
         id = report_result[0][0]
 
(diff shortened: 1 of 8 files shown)


Sparkier:
Not sure if we should require the newest client version with this PR. Technically with the old one can still create breaking projects but the same is true before this commit.

cabreraalex (approved):

cabreraalex:
Hmm could go either way. Annoying to force people to update and very rare bug, but we could be safe

Sparkier:
Since we're mostly hand-holding people rn maybe we don't force them? Getting users without a burden might me more important right now than preventing this very rare bug considering the number of users we have.

cabreraalex:
sold

PR #305, opened 2023-10-13
feat: use slice name as parent folder name when adding all slices

Sparkier:
Fix ZEN-238


cabreraalex:
What happens if there are slices with the same name already?

Sparkier:
We do almost everything by ID now, no? Should not be a problem, right? Also they're all in one folder so no confusion probably.

cabreraalex:
hmm didn't think about that aha, although we prevent people from creating slices with the same name manually.

cabreraalex (approved):

PR #309, opened 2023-10-14
fix: make sure charts always fit into report width

Sparkier:
# Description

Using the [autosize function](https://vega.github.io/vega-lite/docs/size.html#autosize) of Vega.

fix ZEN-239


cabreraalex:
This is epic I wasted so much time trying to figure this out

cabreraalex:
The radar chart is still a bit off, but I think it was off before.

cabreraalex (approved):

PR #311, opened 2023-10-15
fix: fix admin editor on projects

cabreraalex:
# Description
Fix ZEN-250
Fix ZEN-154

### DIFF
--- a/backend/zeno_backend/database/select.py
+++ b/backend/zeno_backend/database/select.py
@@ -896,6 +896,16 @@ def project_state(
         else:
             project.editor = bool(editor[0][0])
 
+        if not project.editor:
+            org_editor = db.execute_return(
+                "SELECT editor FROM organization_project AS op JOIN user_organization "
+                "AS uo ON op.organization_id = uo.organization_id "
+                "WHERE uo.user_id = %s AND op.project_uuid = %s AND uo.admin = TRUE;",
+                [user.id, project_uuid],
+            )
+            if len(org_editor) > 0:
+                project.editor = bool(org_editor[0][0])
+
         return ProjectState(
             project=project,
             metrics=metrics,
--- a/backend/zeno_backend/database/select.py
+++ b/backend/zeno_backend/database/select.py
@@ -542,14 +542,9 @@ def report(
                 "WHERE r.report_id = %s AND o.user_id = %s;",
                 [id, user.id],
             )
-            editor = (
-                bool(org_editor[0])
-                if org_editor is not None and len(org_editor) > 0
-                else False
-            ) or (
-                bool(user_editor)
-                if user_editor is not None and len(user_editor) > 0
-                else False
+            org_editor = [bool(o[0]) for o in org_editor]
+            editor = any(org_editor) or (
+                bool(user_editor[0][0]) if len(user_editor) > 0 else False
             )
 
         linked_projects = db.execute_return(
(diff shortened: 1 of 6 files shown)


backend/zeno_backend/database/select.py
```diff
@@ -884,6 +887,15 @@ def project_state(project_uuid: str, project: Project) -> ProjectState | None:
...
         models = [m[0] for m in model_results]
 
+        editor = db.execute_return(
+            "SELECT editor FROM user_project WHERE user_id = %s AND project_uuid = %s;",
+            [user.id, project_uuid],
+        )
+        if len(editor) == 0:
+            project.editor = False
+        else:
+            project.editor = bool(editor[0][0])
```
Sparkier, line 897: Also need to check org_project I think. Similar to how we do it for the project/report.

Sparkier (changes requested):

backend/zeno_backend/database/select.py
cabreraalex: good catch

backend/zeno_backend/database/select.py
```diff
@@ -884,6 +887,25 @@ def project_state(project_uuid: str, project: Project) -> ProjectState | None:
...
+
+        if not project.editor:
+            org_editor = db.execute_return(
+                "SELECT editor FROM organization_project AS op JOIN user_organization "
+                "AS uo ON op.organization_id = uo.organization_id "
+                "WHERE uo.user_id = %s AND op.project_uuid = %s AND uo.admin = TRUE;",
+                [user.id, project_uuid],
+            )
+            if len(org_editor) > 0:
+                project.editor = bool(org_editor[0][0])
```
Sparkier, line 907: I don't think this is correct. The user does not have to be an admin of the organization to have access and it is sufficient if any of the orgs has edit rights on a project, not necessarily the first one.

Sparkier (changes requested):

backend/zeno_backend/database/select.py
Sparkier: If we do it like this everywhere, I think it's wrong everywhere.

cabreraalex: For (A) this is just for editing, not for viewing check.

For (B) good catch again and I think it might be wrong in other places too.

Sparkier: Even for A, one does not have to be an Org admin to edit if the org has edit rights, correct? Org admin is so you can edit the org settings, not the project settings.

cabreraalex: oh lord okay we should clean up the differences between admin and editor maybe, that makes sense.

Sparkier: Yep, 100%

cabreraalex:
Okay, updated wording from "admin" to editor for most project/report fields.

Sparkier (approved):

backend/zeno_backend/database/select.py
```diff
@@ -884,6 +882,23 @@ def project_state(project_uuid: str, project: Project) -> ProjectState | None:
...
+        user_editor = db.execute_return(
+            "SELECT editor FROM user_project WHERE user_id = %s AND project_uuid = %s",
+            [user.id, project_uuid],
+        )
+        org_editor = db.execute_return(
+            "SELECT editor FROM organization_project AS r JOIN user_organization "
+            "AS o ON r.organization_id = o.organization_id "
+            "WHERE r.project_uuid = %s AND o.user_id = %s;",
+            [project_uuid, user.id],
+        )
```
Sparkier, line 895: We could also do `SELECT * FROM ... WHERE r.editor = True`. Then we would not have to do the `any()`. I assume this would be cheaper and probably more clean.

Sparkier (approved):

PR #318, opened 2023-10-16
feat: more flexible instance view spec

Sparkier:
# Description

Use a specification for instance views so they can be composed from instance elements instead of there being a limited set of fixed instance views.

The current set of instance views is fully supported by this specification. The only exception is that there is now only one chatbot view compared to openai vs. non-openai.

This can be tested using [these notebooks](https://github.com/zeno-ml/projects/tree/main/view_test).

fix ZEN-198
fix ZEN-151

## Views

**Space Separated Values**

image

**Chatbot**

image

**Chatbot Markdown**

image

**Code Generation**

image

**Audio Transcription**

image

**Image Classification**

image

**Text Classification**

image

**RAG**

image


cabreraalex:
woahhh might have you walk me through this looks epic.

Sparkier:
> woahhh might have you walk me through this looks epic.

Yep would probably be good. Maybe I'll also get to some more documentation tomorrow.

frontend/src/lib/components/instances/views/elements/Image.svelte
```diff
@@ -0,0 +1,5 @@
+<script lang="ts">
+	export let data: string;
+</script>
+
+<img src={data} style:max-width="200px" alt="Image thumbnail for instance {data}" />
```
cabreraalex, line 5: Can we make things like the max width a parameter? Are there arbitrary parameter lists options we can pass in to views?

cabreraalex (approved):
Looks great. Biggest overall comment is we need to document this strongly somewhere. Maybe doesn't have to be here but on zenoml.com but should be published at the same time we merge imo

PR #321, opened 2023-10-17
feat: add project names to dropdowns in chart and slice edit and sepa…

Sparkier:
…rate code

Slice and chart selection now includes the project name in brackets. Edit logic has also been separated into different files.

fix ZEN-231

### DIFF
--- a/frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
+++ b/frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
@@ -139,18 +139,22 @@
 			on:consider={handleMoved}
 			on:finalize={handleDropped}
 		>
-			{#each elements as element (element.id)}
-				<ElementContainer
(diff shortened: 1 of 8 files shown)


frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
```diff
@@ -139,18 +139,22 @@
...
-					bind:editId
-					bind:dragEnabled
-					bind:showConfirmDelete
-					{addElement}
-					{chartOptions}
-					{sliceOptions}
-					report={data.report}
-				/>
-			{/each}
+			{#await chartOptions then chartOptions}
```
Comment (cabreraalex) [inline frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:142]: Do we always want to have to await these even if we are just viewing? Just viewing we don't need to make these fetch calls.

cabreraalex (approved):

PR #323, opened 2023-10-17
fix: report page and slice element header

cabreraalex:
# Description
Fix SQL for report home page and fix alignment on slice element headers in reports.

Fix ZEN-252

### DIFF
--- a/frontend/src/lib/components/report/elements/SliceElement.svelte
+++ b/frontend/src/lib/components/report/elements/SliceElement.svelte
@@ -61,25 +61,23 @@
 
 {#if sliceElementOptions !== undefined && sliceElementSpec && table}
 	<div class="w-full">
-		<div class="flex items-center mb-2 justify-between">
-			<div class="flex">
-				<h3 class="text-lg">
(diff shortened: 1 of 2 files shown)


frontend/src/lib/components/report/elements/SliceElement.svelte
```diff
@@ -61,23 +61,25 @@
...
-			<Button
-				variant="outlined"
-				class="ml-4"
-				on:click={() =>
-					goto(
-						`/project/${sliceElementOptions?.project.ownerName}/${sliceElementOptions?.project.name}` +
-							encodeParams()
-					)}>Explore</Button
-			>
+		<div class="flex items-center mb-2 justify-between">
```
Sparkier, line 64: I think I already fixed this in my PR. Don't think this extra div is needed.

Sparkier (approved):

frontend/src/lib/components/report/elements/SliceElement.svelte
cabreraalex: Oop yeah hadn't pulled

PR #334, opened 2023-10-21
feat: redo grid homescreen & order by likes

cabreraalex:
# Description
A couple of things

* Adds pagination to reports and projects on homescreen for scalability
* Combine reports and projects into one homepage. This makes it more discoverable and is possible with the pagination
* Sorts the homescreen by likes
* Refactors server calls to enable pagination & sorting.

Fix ZEN-262

Screenshot 2023-10-20 at 17 08 19


backend/zeno_backend/classes/homepage.py
```diff
@@ -0,0 +1,21 @@
...
+        projects (list[Project]): the projects to be displayed.
+        reports (list[Report]): the reports to be displayed.
+        num_projects (int): the number of projects.
+        num_reports (int): the number of reports.
+    """
+
+    projects: list[Project]
+    reports: list[Report]
+    num_projects: int
+    num_reports: int
```
Sparkier, line 21: I don't think I would fetch this as one. We're saving one request but will have to refactor big time if we ever change/disentangle this again.

backend/zeno_backend/server.py
```diff
@@ -468,97 +445,64 @@ def get_project_uuid(owner_name: str, project_name: str, request: Request):
...
+    @api_app.get("/homepage/", response_model=HomepageData, tags=["zeno"])
+    def get_public_homepage_data(req: Request):
+        projects = select.public_projects(ProjectsRequest(limit=6))
+        reports = select.public_reports(ReportsRequest(limit=6))
+        return HomepageData(
+            projects=projects,
+            reports=reports,
+            num_projects=select.project_count(),
+            num_reports=select.report_count(),
+        )
```
Sparkier, line 489: As said, I would not do this. I don't think we should make our routes match pages but rather objects in the database.

frontend/src/app.css
```diff
@@ -1,3 +1,9 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+.home-grid {
+	display: grid;
+	grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
+	gap: 10px;
+}
```
Sparkier, line 9: Why not use tailwind for this?

backend/zeno_backend/database/select.py
```diff
@@ -32,6 +38,59 @@
...
+    SELECT p.uuid, p.name, p.owner_id, p.view, p.samples_per_page,
+        p.public, p.description, op.editor
+    FROM projects AS p
+    JOIN (SELECT op.project_uuid, uo.organization_id, editor
+            FROM user_organization AS uo
+            JOIN organization_project AS op
+            ON uo.organization_id = op.organization_id
+            WHERE uo.user_id = %s) AS op ON p.uuid = op.project_uuid)
+    """
+)
```
Sparkier, line 67: This is great!

backend/zeno_backend/database/select.py
```diff
@@ -51,115 +110,96 @@ def models(project: str) -> list[str]:
...
+            sql.SQL("WITH LikesSummary AS (")
+            + likes_query
+            + sql.SQL("),")
+            + sql.SQL(" CombinedProjects AS (")
+            + PROJECTS_BASE_QUERY
+            + sql.SQL(") ")
+            + sql.SQL(
+                "SELECT cp.*, COALESCE(ls.total_likes, 0) AS total_likes FROM"
+                " CombinedProjects cp LEFT JOIN LikesSummary ls "
+                " ON cp.uuid = ls.project_uuid ORDER BY total_likes DESC "
```
Sparkier, line 139: This is very helpful.

Sparkier: Something to be considered: What if we want to sort by something different than total number of likes?

Sparkier (changes requested):
High Level Feedback: There are some awesome things in here and some things that we should think more about. It would really help if we could split up this PR to make it smaller contributions, e.g. likes and SQL changes, UI changes (potentially even in steps), ...

Does seem to me like this could be split up reasonably well. As is is very hard to review.

backend/zeno_backend/classes/homepage.py
cabreraalex: So four different requests?

backend/zeno_backend/classes/project.py
```diff
@@ -31,36 +49,7 @@ class Project(CamelModel):
...
-    """Project and details for homepage rendering.
-
-    Attributes:
-        project (Project): project object with project metadata.
-        statistics (ProjectStats): project statistics.
-    """
-
-    project: Project
-    statistics: ProjectStats
+    stats: ProjectStats | None = None
```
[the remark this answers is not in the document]
cabreraalex: Hmm yeah probably not a good idea

backend/zeno_backend/server.py
cabreraalex: I think it's a good bit faster to have a combined request:
https://softwareengineering.stackexchange.com/questions/341872/rest-api-design-multiple-calls-vs-single-call-to-the-api

But not really sure what the right answer is.

frontend/src/app.css
cabreraalex: They didn't have the right property for grid-template-columns... could have this only be one property I guess?

backend/zeno_backend/database/select.py
cabreraalex: Yeah I think future PR will make this an option and use different types of sorts

backend/zeno_backend/classes/homepage.py
cabreraalex: IMO worth not making as many sequential fetch requests we await, makes it significantly slower

frontend/src/app.css
cabreraalex: Updated :)

backend/zeno_backend/classes/project.py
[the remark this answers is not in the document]
cabreraalex: updated!

backend/zeno_backend/classes/homepage.py
Sparkier: I would do two requests and they can be parallel. One for reports, one for projects. I'm pretty sure there will be cases where we only fetch one.

Sparkier:
High level: Since I think you want to disentangle this, let's move the conversation there.

PR #345, opened 2023-10-23
fix: prevent scrolling away the header

Sparkier:
# Description

https://github.com/zeno-ml/zeno-hub/pull/328#discussion_r1367687627


Sparkier:
can we merge this for now @cabreraalex? I think this looks really bad. We can still move to pagination later.

cabreraalex (approved):

PR #349, opened 2023-10-23
fix: update continuous filtering

cabreraalex:
# Description
Fix vega continuous filtering

Fix ZEN-267

### DIFF
--- a/frontend/src/lib/components/metadata/cells/metadata-cells/ContinuousMetadataCell.svelte
+++ b/frontend/src/lib/components/metadata/cells/metadata-cells/ContinuousMetadataCell.svelte
@@ -17,53 +17,47 @@
 	export let updatePredicates: (predicates: FilterPredicate[]) => void;
 
 	let view: View;
-	let localSelection: number[] = [];
-	let localSelectionX: number[] = [];
-
-	function updateSel() {
-		if (filterPredicates.length !== 0) {
-			view.signal('brush', { bucket: localSelection });
-			view.signal('brush_x', localSelectionX);
-		} else {
-			view.signal('brush', {});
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/metadata/cells/metadata-cells/ContinuousMetadataCell.svelte
```diff
@@ -17,57 +17,47 @@
...
+		view.addSignalListener('brush', setSelection);
 	}
 
-	$: if (view) {
-		updateSel(filterPredicates);
-	}
-
-	function setSelection() {
-		if (localSelection.length > 0) {
+	let timer: ReturnType<typeof setTimeout>;
```
Sparkier, line 29: Nit: maybe move to the other lets? Don't know how we do it generally? I normally do exports then vars then reactive then funcs but open to do this differently.

frontend/src/lib/components/metadata/cells/metadata-cells/ContinuousMetadataCell.svelte
```diff
@@ -17,57 +17,47 @@
...
-
-	const endDrag = () => {
-		window.removeEventListener('mouseup', endDrag);
-		setSelection();
-	};
 </script>
 
 <!-- We shallow copy histogram to remove the vega identifiers and force it to update the chart when new data is passed in. -->
-<div id="histogram" on:mousedown={() => window.addEventListener('mouseup', endDrag)}>
+<div id="histogram">
```
Sparkier, line 60: More of a question, do we use this id somehow? Don't really understand how.

Sparkier (approved):

PR #350, opened 2023-10-23
feat: support tabular data

cabreraalex:
# Description
Paired with https://github.com/zeno-ml/zeno-client/compare/ac-tabular?expand=1


Sparkier:
This does not work with the new instance views. I think we should check for an empty spec instead.

PR #351, opened 2023-10-24
refactor: database calls for homepage

cabreraalex:
# Description
Better DB calls for backend, base for pagination on homepage.


backend/zeno_backend/classes/report.py
```diff
@@ -83,16 +83,17 @@ class ReportStats(CamelModel):
...
-class ReportDetails(CamelModel):
-    """Report and details for homepage rendering.
+class ReportsDetails(CamelModel):
+    """Reports and details for homepage rendering.
 
     Attributes:
-        report (Report): report object with report metadata.
-        statistics (ReportStats): report statistics.
+        reports (list[Report]): report object with report metadata.
+        statistics (list[ReportStats]): report statistics.
```
Sparkier, line 91: Attribute missing.

backend/zeno_backend/database/select.py
```diff
@@ -51,115 +102,95 @@ def models(project: str) -> list[str]:
...
-            "WHERE user_id = %s) AS op ON p.uuid = op.project_uuid;",
-            [user.id],
-        )
-        org_projects = []
-        for res in project_org_result:
+        projects = []
+        for res in projects_result:
             owner_name = db.execute_return(
                 "SELECT name FROM users WHERE id = %s", [res[2]]
             )
```
Sparkier, line 146: Why don't we add this to the complex SQL query as well?

backend/zeno_backend/database/select.py
```diff
@@ -51,115 +102,95 @@ def models(project: str) -> list[str]:
...
+            sql.SQL("WITH LikesSummary AS (")
+            + likes_query
+            + sql.SQL("),")
+            + sql.SQL(" CombinedProjects AS (")
+            + PROJECTS_BASE_QUERY
+            + sql.SQL(") ")
+            + sql.SQL(
+                "SELECT cp.*, COALESCE(ls.total_likes, 0) AS total_likes FROM"
+                " CombinedProjects cp LEFT JOIN LikesSummary ls "
+                " ON cp.uuid = ls.project_uuid ORDER BY total_likes DESC "
```
Sparkier, line 131: Can't we make this whole thing the project query? We would have to add likes to the project but maybe we want that anyway...

backend/zeno_backend/database/select.py
```diff
@@ -180,130 +211,92 @@ def public_projects() -> list[Project]:
...
+            + likes_query
+            + sql.SQL("),")
+            + sql.SQL(" CombinedReports AS (")
+            + REPORTS_BASE_QUERY
+            + sql.SQL(") ")
+            + sql.SQL(
+                "SELECT cp.*, COALESCE(ls.total_likes, 0) AS total_likes FROM"
+                " CombinedReports cp LEFT JOIN LikesSummary ls "
+                " ON cp.id = ls.report_id ORDER BY total_likes DESC "
+            )
```
Sparkier, line 241: Same as above.

backend/zeno_backend/database/select.py
```diff
@@ -334,6 +327,55 @@ def public_reports() -> list[Report]:
...
+        int: the number of projects.
+    """
+    db = Database()
+    if user is None:
+        res = db.connect_execute_return(
+            "SELECT COUNT(*) FROM projects WHERE public = TRUE;"
+        )
+    else:
+        res = db.connect_execute_return(
+            sql.SQL("SELECT COUNT(*) FROM") + PROJECTS_BASE_QUERY + sql.SQL(";"),
```
Sparkier, line 346: This might speak against including likes. Don't know.

backend/zeno_backend/server.py
```diff
@@ -513,41 +505,41 @@ def get_projects(current_user=Depends(auth.claim())):
         user = select.user(current_user["username"])
         if user is None:
             return Response(status_code=status.HTTP_401_UNAUTHORIZED)
-        return select.projects(user)
+        return select.projects(user, ProjectsRequest())
```
Sparkier, line 508: Why don't we make this a param of the function as well? What if I want limits here, too?

backend/zeno_backend/server.py
```diff
@@ -558,7 +550,7 @@ def get_reports(current_user=Depends(auth.claim())):
         user = select.user(current_user["username"])
         if user is None:
             return Response(status_code=status.HTTP_401_UNAUTHORIZED)
-        return select.reports(user)
+        return select.reports(user, ReportsRequest())
```
Sparkier, line 553: Same as above.

Sparkier (changes requested):

backend/zeno_backend/database/select.py
cabreraalex: In future we will sort by different things so want to keep it separate. Same for calculating the count don't need to sort?

backend/zeno_backend/database/select.py
cabreraalex: Yeah we can

backend/zeno_backend/database/select.py
cabreraalex: Yup!

cabreraalex:
> Why don't we make this a param of the function as well? What if I want limits here, too? 

We currently don't need them for these calls (from reports) - I think we can add it here if we need to in the future.

Sparkier (approved):

PR #356, opened 2023-10-25
feat: fix report element edit w/o project

cabreraalex:
# Description
Fixes reports without projects showing. Also updates styles as discussed:

Screenshot 2023-10-24 at 17 02 59

### DIFF
--- a/frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
+++ b/frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
@@ -84,7 +84,7 @@
 			</h1>
 		</div>
 		<h5 class="mt-4 ml-1 text-lg">Author: {data.report.ownerName}</h5>
-		<hr class="mt-4 bg-grey-dark" />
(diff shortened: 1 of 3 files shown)


frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
```diff
@@ -84,7 +84,7 @@
 			</h1>
 		</div>
 		<h5 class="mt-4 ml-1 text-lg">Author: {data.report.ownerName}</h5>
-		<hr class="mt-4 bg-grey-dark" />
+		<hr class="mt-4 text-grey-light" />
```
Comment (Sparkier) [inline frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:87]: Should we make this a border on the <p> element below? Like this we have these two lines even if there is no associated project.

Sparkier (approved):

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:87]: The other one is in the component so will only render one.

PR #357, opened 2023-10-25
build: replace black with ruff, remove old packages

cabreraalex:
# Description
Requires new views package including highlight

blocked on https://github.com/zeno-ml/instances/pull/19


.github/workflows/ci.yml
```diff
@@ -25,22 +25,17 @@ jobs:
       - name: Svelte check
         run: cd frontend && PUBLIC_AMPLITUDE_API_KEY="" npm run check
 
-  format-backend:
+  backend:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: psf/black@stable
```
Sparkier, line 32: ooc: why are we doing this?

Sparkier (approved):

backend/pyproject.toml
```diff
@@ -48,6 +42,7 @@ include = ["zeno_backend"]
 
 [tool.ruff]
 select = ["E", "F", "I", "PD", "N", "D", "T20"]
+ignore = ["D206", "E501"]
```
Sparkier, line 45: Why do we ignore these? Esp. line too long seems like something we want in code. Also should probably add comments on what these ignores do.

Sparkier (changes requested):

.github/workflows/ci.yml
cabreraalex: ruff is 10-100x faster and then we only need one package to rule them all:

https://twitter.com/tiangolo/status/1716916271793750027

backend/pyproject.toml
cabreraalex: Did some reading, this will still lint and format long lines, the different specs just have different definitions. Added comment.

neubig (approved):
LGTM, thanks!

PR #360, opened 2023-10-25
fix: fetch instance view for report slices

cabreraalex:
# Description
Fix two issues:
* Check that chartOptions exist before selecting chart
* Fetch instance view from string spec for rendering.


neubig (approved):
Looks reasonable to me!

PR #368, opened 2023-10-26
fix: hotfix beeswarm and radar sizes

Sparkier:
# Description

Not a perfect solution but at least makes them work again.

fix ZEN-249


cabreraalex:
MVP

cabreraalex (approved):

PR #369, opened 2023-10-26
fix: various fixes

cabreraalex:
# Description
* Add two options to use yalc for locally using instances package: https://stackoverflow.com/questions/68670700/npm-link-dev-packages-when-using-docker-dev-containers
* Fix ZEN-295
* Fix ZEN-291 - Use grid view in instances and fixed half-width in reports for instance view to fill space. Eventually want to make it more customizable, for now it's max width of 400.
* Fix ZEN-294 - validate range and set on enter.
* Fix ZEN-289

### DIFF
--- a/frontend/src/lib/components/instances/ListView.svelte
+++ b/frontend/src/lib/components/instances/ListView.svelte
@@ -97,7 +97,10 @@
 	}
 </script>
 
-<div class="overflow-y-auto flex flex-wrap content-start w-full h-full">
+<div
+	class="overflow-y-auto w-full h-full grid"
+	style="grid-template-columns: repeat(auto-fill, minmax(400px, 1fr))"
+>
 	{#await tablePromise then table}
 		{#if idColumn !== undefined}
 			{#each table as inst (inst[idColumn])}
--- a/frontend/src/lib/components/instances/ListView.svelte
+++ b/frontend/src/lib/components/instances/ListView.svelte
@@ -99,7 +99,7 @@
 
 <div
 	class="overflow-y-auto w-full h-full grid"
-	style="grid-template-columns: repeat(auto-fill, minmax(400px, 1fr))"
+	style="grid-template-columns: repeat(auto-fill, minmax(400px, 1fr)); grid-auto-rows: min-content;"
 >
 	{#await tablePromise then table}
 		{#if idColumn !== undefined}
--- a/frontend/src/lib/components/metadata/MetricRange.svelte
+++ b/frontend/src/lib/components/metadata/MetricRange.svelte
@@ -1,5 +1,20 @@
 <script lang="ts">
 	import { metric, metricRange, metrics } from '$lib/stores';
+
+	function updateRange(e: Event, i: number) {
+		const target = e.target as HTMLElement;
+
+		let val = parseFloat(target.innerText);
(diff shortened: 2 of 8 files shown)


frontend/src/lib/components/instances/ListView.svelte
```diff
@@ -97,7 +97,10 @@
 	}
 </script>
 
-<div class="overflow-y-auto flex flex-wrap content-start w-full h-full">
+<div
+	class="overflow-y-auto w-full h-full grid"
+	style="grid-template-columns: repeat(auto-fill, minmax(400px, 1fr))"
+>
```
Sparkier, line 103: Looks awesome on my big screen, but on my integrated one somewhat off:
image

1. It could get a bit wider to fill the remaining blank space, no? Not a grid expert, though.
2. The vertical whitespace looks pretty bad.

frontend/src/lib/components/metadata/MetricRange.svelte
```diff
@@ -9,24 +24,34 @@
 		</span>
 		<span
 			contenteditable="true"
-			on:blur={(e) =>
-				metricRange.update((range) => {
-					range[0] = parseFloat(e.currentTarget.innerText);
-					return [...range];
-				})}
+			role="textbox"
+			aria-label="Metric range min"
```
Sparkier, line 28: ❤️

frontend/src/lib/components/metadata/MetricRange.svelte
```diff
@@ -9,24 +24,34 @@
...
+					e.currentTarget.blur();
+				}
+			}}
+			on:blur={(e) => updateRange(e, 1)}
 		>
 			{$metricRange[0].toFixed(2)}
 		</span>
 		<div class="w-10 h-4 mx-2.5 bg-gradient-to-r from-primary-light to-primary" />
-		<span
+		<div
```
Sparkier, line 41: nit: why is this a div now when the other one is still a span?

frontend/src/lib/components/metadata/MetricRange.svelte
```diff
@@ -1,5 +1,20 @@
 <script lang="ts">
 	import { metric, metricRange, metrics } from '$lib/stores';
+
+	function updateRange(e: Event, i: number) {
```
Sparkier, line 4: Nit: Variable name for `i` is not super self-explanatory. If we change this and potentially add a function doctoring maybe we can make it easier to understand this function.

Sparkier (changes requested):

frontend/src/lib/components/instances/ListView.svelte
Sparkier: I noticed that 1 might be because I don't have your version of instance views.

frontend/src/lib/components/metadata/MetricRange.svelte
cabreraalex: good catch

cabreraalex:
Okay two updates:
* Fixed the grid rows so they don't fill the space
* Fixed the comparison view table so it doesn't overflow like crazy

Sparkier (approved):
Make sure to update the package version for instance views before merging.

PR #371, opened 2023-10-26
feat: grid layout for homescreen

cabreraalex:
# Description

Use a grid layout to size homescreen, and remove grouping of projects/reports

### DIFF
--- a/frontend/.yalc/@zeno-ml/zeno-instance-views/yalc.sig
+++ b/frontend/.yalc/@zeno-ml/zeno-instance-views/yalc.sig
@@ -1,1 +1,0 @@
-91a80a073f5bedfca0f5050ddbfe7981
--- a/frontend/.yalc/@zeno-ml/zeno-instance-views/yalc.sig
+++ b/frontend/.yalc/@zeno-ml/zeno-instance-views/yalc.sig
@@ -1,0 +1,1 @@
+91a80a073f5bedfca0f5050ddbfe7981
--- a/frontend/src/lib/components/report/Report.svelte
+++ b/frontend/src/lib/components/report/Report.svelte
@@ -42,7 +42,7 @@
 	on:focus={() => (hovering = true)}
 	on:mouseleave={() => (hovering = false)}
 	on:blur={() => (hovering = false)}
-	class="border-solid mr-2 mb-2 rounded-lg border-grey-light border shadow-sm flex flex-col p-3 px-5 hover:shadow-md w-full sm:w-80 h-60"
+	class="border-solid rounded-lg border-grey-light border shadow-sm flex flex-col p-3 px-5 hover:shadow-md h-full"
 >
 	<div class="flex justify-between w-full">
 		<div class={deletable ? 'mr-5' : ''}>
--- a/frontend/src/lib/components/report/Report.svelte
+++ b/frontend/src/lib/components/report/Report.svelte
@@ -90,8 +90,13 @@
 			{/if}
 		</div>
 	</div>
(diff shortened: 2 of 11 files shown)


frontend/.yalc/@zeno-ml/zeno-instance-views/yalc.sig
```diff
@@ -0,0 +1 @@
+91a80a073f5bedfca0f5050ddbfe7981
```
Sparkier, line 1: Probably needs the other PR for this to be ignored, correct?

frontend/src/lib/components/report/Report.svelte
```diff
@@ -43,7 +43,7 @@
 	on:focus={() => (hovering = true)}
 	on:mouseleave={() => (hovering = false)}
 	on:blur={() => (hovering = false)}
-	class="border-solid mr-2 mb-2 rounded-lg border-grey-light border shadow-sm flex flex-col p-3 px-5 hover:shadow-md w-full sm:w-80 h-60"
+	class="border-solid rounded-lg border-grey-light border shadow-sm flex flex-col p-3 px-5 hover:shadow-md h-full"
```
Sparkier, line 46: Moving forward, maybe we could make this card a component so that we don't duplicate code between reports and projects.

frontend/src/routes/(app)/(home)/+layout.svelte
```diff
@@ -9,7 +9,7 @@
 		$page.route.id === '/(app)/(home)/reports' || $page.route.id === '/(app)/(home)/projects';
 </script>
 
-<div class="flex flex-col w-full h-full p-8 pt-5 bg-white">
+<div class="flex flex-col w-full h-full p-8 pt-5 bg-white max-w-7xl m-auto">
```
Sparkier, line 12: Can't we keep it full width with the grid?

Sparkier (changes requested):
Looks like this on all types of screens for me:

image

I assume this is not intended.

frontend/.yalc/@zeno-ml/zeno-instance-views/yalc.sig
cabreraalex: oops yeah removed it

cabreraalex:
odd the tailwind didn't get pushed. Fixed a couple of things
* made the card container a component (not the one you mentioned though, that can wait I think)
* Limited description size in cards

As for limiting widht, I think it looks bad and isn't very usable if we just let it extend horizontally forever. Having a max width and having it go vertically I think is more ergonomic even on bigger screens.

Sparkier:
> odd the tailwind didn't get pushed. Fixed a couple of things
> 
> * made the card container a component (not the one you mentioned though, that can wait I think)
> * Limited description size in cards
> 
> As for limiting widht, I think it looks bad and isn't very usable if we just let it extend horizontally forever. Having a max width and having it go vertically I think is more ergonomic even on bigger screens.

Maybe don't scale indefinitely, but this definitely looks bad:

image

Sparkier (changes requested):
Still not convinced by the width and it should not show the scroll bars if they are not needed (and if, only overflow-y).

cabreraalex:
I'm don't feel that strongly, removed the max width

Sparkier:
Scroll bars still broken for me.  Can try to fix on my end tomorrow if you don't see this: 

image

Sparkier:
It's the tooltips... Need to test a bit more how to fix but if you remove them the scroll bars are gone.

Sparkier (approved):

PR #373, opened 2023-10-27
feat: add checkbox to select all models/metrics/slices for a chart

Sparkier:
# Description

fix ZEN-204

### DIFF
--- a/backend/zeno_backend/processing/chart.py
+++ b/backend/zeno_backend/processing/chart.py
@@ -1,6 +1,5 @@
 """Functions for extracting chart data from SQL."""
 import json
-from functools import lru_cache
 from typing import Any
 
 from zeno_backend.classes.chart import (
@@ -80,7 +79,7 @@ def get_selected_models(chart_models: list[str], project: str) -> list[str]:
     Returns:
         list[str]: list of models of the chart.
     """
-    if "all_zeno_project_models" in chart_models:  # all models should be selected
+    if "" in chart_models:  # all models should be selected
         return models(project)
     return chart_models
 
@@ -336,7 +335,6 @@ def heatmap_data(chart: Chart, project: str) -> str:
     return json.dumps({"table": elements})
 
 
-@lru_cache(4096)
 def chart_data(chart: Chart, project: str) -> str:
     """Extract the chart data for a specific chart that the user created.
 
--- a/backend/zeno_backend/processing/chart.py
+++ b/backend/zeno_backend/processing/chart.py
@@ -17,9 +17,72 @@ from zeno_backend.classes.chart import (
 from zeno_backend.classes.filter import FilterPredicateGroup, Join
 from zeno_backend.classes.metric import Metric
 from zeno_backend.classes.slice import Slice
-from zeno_backend.database.select import metrics, slices
+from zeno_backend.database.select import metrics, models, slices
 from zeno_backend.processing.filtering import table_filter
 from zeno_backend.processing.metrics.map import metric_map
+
+
+def get_selected_slices(chart_slices: list[int], project: str) -> list[Slice]:
+    """Get the slices of the chart from a list of slice ids.
+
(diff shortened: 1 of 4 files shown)


cabreraalex:
We need to get rid of the @lru_cache now since it won't update otherwise

backend/zeno_backend/processing/chart.py
```diff
@@ -17,11 +17,74 @@
...
+    """Get the slices of the chart from a list of slice ids.
+
+    Args:
+        chart_slices (list[int]): slice ids of the chart.
+        project (str): project the user is currently working with.
+
+    Returns:
+        list[Slice]: list of slices of the chart.
+    """
+    if -2 in chart_slices:  # -2 signals that all slices should be selected
```
cabreraalex, line 35: Should this include `all_instances`? I can go either way.

backend/zeno_backend/processing/chart.py
```diff
@@ -17,11 +17,74 @@
...
+    """Get the models of the chart from a list of models.
+
+    Args:
+        chart_models (list[str]): the list of models of the chart.
+        project (str): project the user is currently working with.
+
+    Returns:
+        list[str]: list of models of the chart.
+    """
+    if "all_zeno_project_models" in chart_models:  # all models should be selected
```
cabreraalex, line 83: what if someone names a model this ahaha

frontend/src/lib/components/chart/chart-page/encoding/chart-encoding/encoding-components/ModelsEncodingMultiChoice.svelte
```diff
@@ -27,12 +28,26 @@
...
-	/>
+	<div class="flex flex-col">
+		{#if value[0] != 'all_zeno_project_models'}
+			<Svelecte
+				style="width: 280px; flex:none;"
+				bind:value
+				{options}
+				{dndzone}
+				multiple
+				placeholder="Select Slices..."
```
cabreraalex, line 39: "Systems..."

frontend/src/lib/components/chart/chart-page/encoding/chart-encoding/encoding-components/ModelsEncodingMultiChoice.svelte
```diff
@@ -27,12 +28,26 @@
...
+			/>
+		{/if}
+		<div class="flex items-center ml-auto">
+			<span>All Models</span>
```
cabreraalex, line 43: "All Systems"

cabreraalex (changes requested):

backend/zeno_backend/processing/chart.py
Sparkier: Don't know either.... Often destroys the scales that's why I did not include it.

backend/zeno_backend/processing/chart.py
Sparkier: Yep that's a problem. The alternative would be adding another field but don't really like that either. Don't know what to do....

backend/zeno_backend/processing/chart.py
cabreraalex: Yeah I'm alright not including it.

backend/zeno_backend/processing/chart.py
cabreraalex: does an empty string work?

cabreraalex:
Can you also remove the @lru_cache

cabreraalex (approved):

backend/zeno_backend/processing/chart.py
cabreraalex: if not it's fine super edge case

Sparkier: Very good idea!

PR #378, opened 2023-10-30
refactor: implement custom tooltip to fix errors with tooltip library

Sparkier:
# Description

fix ZEN-305

Implements a custom tooltip that uses the `use:` directive to listen to mouse events and uses a global tooltip state and element.


cabreraalex:
Looks great but check linting error

cabreraalex (approved):

cabreraalex:
This has a small issue - if you are hovering on something with a tooltip and the window changes the tooltip gets stuck.

e.g. to reproduce hover on the number of systems on a card and click to go in.

PR #382, opened 2023-11-01
build: tailwind autosort

cabreraalex:
# Description

Automatically sort tailwind classes


cabreraalex:
Having some issues with vscode and the npm commands doing different things

.vscode/settings.json
```diff
@@ -1,12 +1,11 @@
 {
   "python.defaultInterpreterPath": ".venv/bin/python",
-  "autoDocstring.docstringFormat": "google",
```
Sparkier, line 3: Shouldn't we keep this?

frontend/.prettierrc
```diff
@@ -3,13 +3,5 @@
...
 	"printWidth": 100,
-	"plugins": ["prettier-plugin-svelte"],
-	"overrides": [
-		{
-			"files": "*.svelte",
-			"options": {
-				"parser": "svelte"
-			}
-		}
-	]
```
Sparkier, line 14: Sure this is not needed?

frontend/package.json
```diff
@@ -40,7 +40,7 @@
 		"eslint-plugin-svelte": "^2.33.0",
 		"postcss": "^8.4.29",
 		"prettier": "^3.0.3",
-		"prettier-plugin-svelte": "^3.0.3",
+		"prettier-plugin-tailwindcss": "^0.5.6",
```
Sparkier, line 43: Don't we need both plugins?

frontend/src/app.postcss
```diff
@@ -1,3 +1,6 @@
+/* Write your global styles here, in PostCSS syntax */
+
+@tailwind base;
```
Sparkier, line 3: What's going on here?

Sparkier (changes requested):

frontend/package.json
cabreraalex: the svelte one comes bundled in this one

.vscode/settings.json
cabreraalex: I stopped using it and just use copilot so it showed not used but yes can keep probably better.

frontend/src/app.postcss
cabreraalex: I think this is supposed to be postcss, I re-ran the setup for tailwind. Also updated import in main app component.

frontend/.prettierrc
cabreraalex: it was added back

cabreraalex:
This is working consistently across VSCode, CLI, and GH Actions for me

Sparkier (approved):

PR #383, opened 2023-11-01
feat: show update date on report and creation and update date in proj…

Sparkier:
…ect settings

# Description

fix ZEN-307

image
image

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -42,8 +42,6 @@ include = ["zeno_backend"]
 
 [tool.ruff]
 select = ["E", "F", "I", "PD", "N", "D", "T20"]
-# Ignore conflicting specs between linter and formatter. Formatter will catch line length.
-ignore = ["D206", "E501"]
 
 [tool.ruff.pydocstyle]
 convention = "google"
--- a/frontend/src/lib/components/popups/ProjectPopup.svelte
+++ b/frontend/src/lib/components/popups/ProjectPopup.svelte
@@ -109,8 +109,8 @@
 						<Checkbox checked={config.public} on:click={() => (config.public = !config.public)} />
 						<span>Public visibility</span>
 					</div>
-					<span>Created at: {new Date(config.createdAt).toLocaleDateString()}</span>
-					<span>Last updated at: {new Date(config.updatedAt).toLocaleDateString()}</span>
+					<span>Created: {new Date(config.createdAt).toLocaleString()}</span>
+					<span>Updated: {new Date(config.updatedAt).toLocaleString()}</span>
 				</div>
 			</div>
 			<div>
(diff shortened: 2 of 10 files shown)


backend/pyproject.toml
```diff
@@ -42,8 +42,6 @@ include = ["zeno_backend"]
 
 [tool.ruff]
 select = ["E", "F", "I", "PD", "N", "D", "T20"]
-# Ignore conflicting specs between linter and formatter. Formatter will catch line length.
-ignore = ["D206", "E501"]
```
Sparkier, line 46: This has to be removed otherwise the CI does not catch things that should be caught (e.g. long lines)

cabreraalex:
Should we show the time too? It then gets iffy on time zones but might be useful to see time.

Sparkier:
> Should we show the time too? It then gets iffy on time zones but might be useful to see time.

Maybe in the settings, I would not on the report page, not really relevant there, is it?

cabreraalex:
Can collaborators see the settings? asking for the class assignment for the profs to see the time maybe.

frontend/src/lib/components/popups/ProjectPopup.svelte
```diff
@@ -109,6 +109,8 @@
 						<Checkbox checked={config.public} on:click={() => (config.public = !config.public)} />
 						<span>Public visibility</span>
 					</div>
+					<span>Created at: {new Date(config.createdAt).toLocaleDateString()}</span>
```
cabreraalex, line 112: I think just `Created` and `Updated` is fine

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
```diff
@@ -91,11 +91,13 @@
 				report
 			/>
 		</div>
-		<h5 class="mt-4 ml-1 text-lg">Author: {data.report.ownerName}</h5>
-		<hr class="mt-4 text-grey-light" />
+		<h5 class="mt-4 text-lg">Author: {data.report.ownerName}</h5>
+		<span class="text-grey-darker mt-2">{new Date(data.report.updatedAt).toLocaleDateString()}</span
```
Comment (cabreraalex) [inline frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:95]: I would add `Updated` before the date here. no colon needed just `Update 10/10/2023 at 12:23`

cabreraalex (changes requested):
Do we want to show this on projects too?

Sparkier:
> Do we want to show this on projects too?

I don't think it's as important for project and I'm not sure where. But could show as well no strong opinion.

cabreraalex:
Fine no for no

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
Comment (Sparkier) [reply frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:95]: Actually, maybe fine with time generally:
image

cabreraalex (approved):

PR #384, opened 2023-11-01
feat: add new project button that links to docs

Sparkier:
# Description

Not sure if we really want this...

image

fix ZEN-311


cabreraalex:
If we do this I would make it a modal that pops up with some version of the guide on how to use the API. But can go either way.

Sparkier:
Maybe we don't need this button since if you don't have any projects yet we're showing this and if you've already created one you should know how to create more.

Probably rather focus more on improving guidance for the first one.

cabreraalex:
SGTM

PR #390, opened 2023-11-02
feat: enable filtering for individual instance

Sparkier:
# Description

fix ZEN-166
fix ZEN-341

image


cabreraalex:
This is awesome, two high-level things:
* I would design the selection interaction slightly differently - when you hover on an instance we maybe make the border purple and show a circular check in the top right. When you click the check, then we make the border purple and make it part of the selection? Maybe then people can select a few instances and click an "update selection" button that _then_ updates the selection?

* Relatedly, we should think how it mentally connects to tags and potentially the projection/scatterplot. In the scatterplot we had a concept of a "current selection"  similar to the above we could maybe reuse? We also have checkboxes on the table view we should maybe always show to allow for selection?

Anyway maybe worth chatting about and maybe ask designers for input too

Sparkier:
> This is awesome, two high-level things:
> 
> * I would design the selection interaction slightly differently - when you hover on an instance we maybe make the border purple and show a circular check in the top right. When you click the check, then we make the border purple and make it part of the selection? Maybe then people can select a few instances and click an "update selection" button that _then_ updates the selection?
> * Relatedly, we should think how it mentally connects to tags and potentially the projection/scatterplot. In the scatterplot we had a concept of a "current selection"  similar to the above we could maybe reuse? We also have checkboxes on the table view we should maybe always show to allow for selection?
> 
> Anyway maybe worth chatting about and maybe ask designers for input too

This all makes sense to me. I agree we should talk about how and where selection works. I implemented this as more of a hot fix to unblock one of the projects but it seems like that project is not blocked by this anymore. Thus, happy to talk about and iterate before merging.

cabreraalex:
Ways to select your data
* Slices -> combinations of metadata filters
* Tag -> Set of IDs
* Current selection -> Set of IDs

Combine all of these -> slice + tag + selected IDs -> took intersection of those.

Sparkier:
Completely changed now @cabreraalex to reflect what we've been talking about.

I am not really convinced by the color scheme (specifically the when to put green vs purple and by the green color in the first place).

cabreraalex:
High-level comments:
* We should show the checkmark without hover when something is selected
* The hover circle of the checkmark is overflowing on the right side of the instance.
* The checkmark position likely interferes with the image expand location - should change the position of one or the other
* Clicking the "4 instances selected" to create a tag is very undiscoverable (only found it by accident) - curious what should be here in general... Maybe we just show a button that says [Create tag with X instance] and an (x) next to it to clear? 
* Color-wise I kind of like the teal honestly aha.

cabreraalex (changes requested):
Left high-level comments

Sparkier:
good catches. fixed them and made it more like the filter bar in terms of style.

cabreraalex (changes requested):
Sweeeet. A few more thoughts:

When editing a tag the interface is a bit confusing. We should definitely update the text to say "update tag" instead of create tag. We should also get rid of the "Editing" tag name in the top right, and not automatically switch to table view and keep the view picker since we can now add things in the main view. Also maybe get rid of the "done" button next to the tag?

cabreraalex:
We should also think about what selection should be like in the comparison view. Right now if I select a few things and go  back to the explore page I don't know how to look at what I selected. Maybe if you click "N selected items" it filters for the selected items and turns green? And click to unfilter for them?

Sparkier:
Very good ideas. All implemented.

https://github.com/zeno-ml/zeno-hub/assets/5690524/939e48e6-0d17-4445-aeef-e72c0bf76b82

cabreraalex (approved):
three last comments for tags, it looks awesome.
* Can we remove the await block and use a callback to avoid the flickering of the metric number next to the tag?
* Can we change the button to say “update” or “update tag” when editing a tag?
* When clicking edit or clicking an element the sorting sometimes changes and I lose the element I was looking for… idk why or if that can be fixed

PR #391, opened 2023-11-02
feat: add embedding metadata type

cabreraalex:
# Description
Add embedding as a column type for later work on similarity search, etc.

### DIFF
--- a/frontend/src/lib/util/util.ts
+++ b/frontend/src/lib/util/util.ts
@@ -42,18 +42,16 @@ export function getProjectRouteFromURL(url: URL) {
 
 function columnTypeOrder(colType: ZenoColumnType) {
 	switch (colType) {
-		case ZenoColumnType.FEATURE:
+		case ZenoColumnType.ID:
 			return 0;
-		case ZenoColumnType.OUTPUT:
-			return 1;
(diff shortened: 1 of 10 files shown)


cabreraalex:
We likely want to also have a PROJECTION type that is 2D numeric vectors - it's harder to detect this though, no explicit type in the arrow schema. Would have to wait to have the data later and query the DB, do some inference.

cabreraalex:
Could also do this dynamically where when you go to projection tab we get all the embedding cols and see which are 2D

Sparkier:
> Could also do this dynamically where when you go to projection tab we get all the embedding cols and see which are 2D

I like this idea.

frontend/src/lib/util/util.ts
```diff
@@ -42,18 +42,16 @@ export function getProjectRouteFromURL(url: URL) {
 
 function columnTypeOrder(colType: ZenoColumnType) {
 	switch (colType) {
-		case ZenoColumnType.FEATURE:
+		case ZenoColumnType.ID:
```
Sparkier, line 45: Don't think this makes sense in many cases. Used this for putting feature columns up top in the selects.

Sparkier (approved):

PR #392, opened 2023-11-03
feat: add embeddable featured elements

Sparkier:
# Description

can be used as follows:

https://github.com/zeno-ml/zeno-hub/assets/5690524/ed176666-6c9e-4723-97e9-059b9fa7d35d

fix ZEN-314

### DIFF
--- a/frontend/src/lib/components/home/HomeCard.svelte
+++ b/frontend/src/lib/components/home/HomeCard.svelte
@@ -75,7 +75,7 @@
 	on:focus={() => (hovering = true)}
 	on:mouseleave={() => (hovering = false)}
 	on:blur={() => (hovering = false)}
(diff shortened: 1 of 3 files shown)


cabreraalex:
do we want to make each card a single IFrame and people can combine them? Or add a URL parameter to specify how many are shown? Also maybe change the folder from `featured` to `entry` or something

Sparkier:
> do we want to make each card a single IFrame and people can combine them? Or add a URL parameter to specify how many are shown? Also maybe change the folder from `featured` to `entry` or something

Added a param to specify the number of items. I put it into `entry` but kept the `featured` folder. Might have individual embeddable entries as well at some point.

Sparkier:
You need to leave a review @cabreraalex for me to be able to re-request a review from you.

PR #395, opened 2023-11-03
fix: character rendering for svelecte

cabreraalex:
# Description
Fix ZEN-322

A bit of a hot-fix, I don't know why special characters get rendered so oddly by svelecte.

### DIFF
--- a/frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
+++ b/frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
@@ -61,6 +61,10 @@
 	function handleMoved(e: CustomEvent) {
 		elements = e.detail.items;
 	}
+
+	function optionsRenderer(o: { name: string }) {
+		return o.name;
+	}
 </script>
 
 {#if showConfirmDelete !== -1}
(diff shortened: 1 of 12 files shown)


cabreraalex:
Yeah this is annoying it's hardcoded in the default formatter to escape HTML characters like < >

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
```diff
@@ -61,6 +61,10 @@
 	function handleMoved(e: CustomEvent) {
 		elements = e.detail.items;
 	}
+
+	function optionsRenderer(o: { name: string }) {
+		return o.name;
+	}
```
Comment (Sparkier) [inline frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:67]: This can't be an arrow function because of types I assume?

Comment (Sparkier) [reply frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:67]: Should we make this a util function as well? Used in many places. Don't know...

Sparkier (commented):
Kinda bad that we have to do this... Guess there's no way around, though.

Sparkier (approved):

PR #396, opened 2023-11-03
fix: comparative slice finder

cabreraalex:
# Description
Fixes comparative slice finder for second option

Fix ZEN-324

### DIFF
--- a/backend/zeno_backend/processing/util.py
+++ b/backend/zeno_backend/processing/util.py
@@ -47,8 +47,10 @@ def generate_diff_cols(
     if diff_col_1.data_type == MetadataType.CONTINUOUS:
         if order_by == "descending":
             df.loc[:, "diff"] = df[col1_id] - df[col2_id]
-        else:
+        elif order_by == "ascending":
             df.loc[:, "diff"] = df[col2_id] - df[col1_id]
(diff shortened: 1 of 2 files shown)


backend/zeno_backend/processing/util.py
```diff
@@ -40,7 +45,10 @@ def generate_diff_cols(
 
     # various metadata type difference
     if diff_col_1.data_type == MetadataType.CONTINUOUS:
-        df.loc[:, "diff"] = df[col1_id] - df[col2_id]
+        if order_by == "descending":
+            df.loc[:, "diff"] = df[col1_id] - df[col2_id]
+        else:
+            df.loc[:, "diff"] = df[col2_id] - df[col1_id]
```
neubig, line 51: Looks good but we should either do this, or add a check at the beginning of the function to make sure that the value of `order_by` is legal.
```suggestion
        elif order_by == "ascending":
            df.loc[:, "diff"] = df[col2_id] - df[col1_id]
        else:
            raise ValueError(f"Illegal value for {order_by=}")
```

neubig (changes requested):

neubig (approved):

PR #409, opened 2023-11-06
feat: export reports as PDF

Sparkier:
# Description

fix ZEN-278

https://github.com/zeno-ml/zeno-hub/assets/5690524/c684ade3-844f-44c4-8373-9152a811bb68

The PDF output is not always super clean. E.g. Vega charts sometimes to weird things with sizing. Do we generally want this?


cabreraalex:
Yeah not sure if it's worth given the quality of the PDF...

PR #411, opened 2023-11-06
fix: fix overflow on big table

cabreraalex:
# Description

Fixes overflow on large tables.

I couldn't figure out the right dynamic way to do it and wasted a couple of hours so just hardcoded.

### DIFF
--- a/frontend/src/routes/(app)/project/[owner]/[project]/[page=exploreOrCompare]/+page.svelte
+++ b/frontend/src/routes/(app)/project/[owner]/[project]/[page=exploreOrCompare]/+page.svelte
@@ -33,16 +33,16 @@
 		</Banner>
 	</div>
 {:else}
-	<div class="flex h-full min-h-0 min-w-0 flex-grow flex-col">
+	<div class="flex h-full min-h-0 w-full min-w-0 flex-grow flex-col">
 		<ProjectHeader
 			project={data.project}
 			likes={data.numLikes}
 			liked={data.userLiked}
 			user={data.user}
 		/>
-		<div class="flex h-full min-h-0">
+		<div class="flex h-full min-h-0 w-full min-w-0">
 			<MetadataPanel compare={data.compare} />
-			<div class="mx-5 flex flex-grow flex-col">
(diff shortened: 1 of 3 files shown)


frontend/src/routes/(app)/project/[owner]/[project]/[page=exploreOrCompare]/+page.svelte
```diff
@@ -33,16 +33,16 @@
...
 			project={data.project}
 			likes={data.numLikes}
 			liked={data.userLiked}
 			user={data.user}
 		/>
-		<div class="flex h-full min-h-0">
+		<div class="flex h-full min-h-0 w-full min-w-0">
 			<MetadataPanel compare={data.compare} />
-			<div class="mx-5 flex flex-grow flex-col">
+			<div class="flex w-full flex-grow flex-col px-3" style="max-width: calc(100% - 380px);">
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[owner]/[project]/[page=exploreOrCompare]/+page.svelte:45]: Can we try to avoid these absolute pixels values here? Will break if we ever change the width of the Panel or side header. This should be possible without the calc.

Sparkier (changes requested):

frontend/src/routes/(app)/project/[owner]/[project]/[page=exploreOrCompare]/+page.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/project/[owner]/[project]/[page=exploreOrCompare]/+page.svelte:45]: I tried but failed epically

Comment (cabreraalex) [reply frontend/src/routes/(app)/project/[owner]/[project]/[page=exploreOrCompare]/+page.svelte:45]: Yeah spent another hour or two on this and couldn't figure it out. I'm okay merging like this unless you want to give it a go.

This is breaking table currently.

Comment (Sparkier) [reply frontend/src/routes/(app)/project/[owner]/[project]/[page=exploreOrCompare]/+page.svelte:45]: Let's get this fix in I can try my luck later this week: https://linear.app/zenoml/issue/ZEN-327/remove-hard-coded-width

Sparkier (approved):

PR #416, opened 2023-11-09
feat: updated header styles

cabreraalex:
# Description
Updated headers. Lots of changes on all pages so worth running locally.


Sparkier:
Maybe we should leave this in dev until Monday to make sure we're not breaking things.

PR #417, opened 2023-11-09
feat: fix small report issues, add single-block move

cabreraalex:
# Description

* Fix overflow on charts
* Center spinner
* Fix issue where position was set wrong on new blocks
* Add one up/one down buttons to move without drag

### DIFF
--- a/frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
+++ b/frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
@@ -62,6 +62,23 @@
 	function handleMoved(e: CustomEvent) {
 		elements = e.detail.items;
 	}
+
+	function updateElementPosition(elementId: number | null | undefined, position: number) {
+		if (!elementId || position < 0 || position >= elements.length) return;
+
+		let currElement = elements.find((e) => e.id === elementId);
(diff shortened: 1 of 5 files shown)


frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
```diff
@@ -62,6 +62,23 @@
 	function handleMoved(e: CustomEvent) {
 		elements = e.detail.items;
 	}
+
+	function updateElementPosition(elementId: number | null | undefined, position: number) {
```
Comment (Sparkier) [inline frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:66]: Since this just swaps positions, maybe name it that?

Comment (Sparkier) [reply frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte:66]: Was confused at first and might lead one to think they can move an element to a new position instead of swap the position with another element.

Sparkier (changes requested):

frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/report/[owner]/[report]/+page.svelte]: SGTM!

Sparkier (approved):

PR #418, opened 2023-11-09
refactor: move instance-views into hub

cabreraalex:
# Description

Instead of having a separate instance-views repo, moves it into zeno-hub and creates a playground endpoint.

### DIFF
--- a/frontend/src/routes/playground/+layout.ts
+++ b/frontend/src/routes/playground/+layout.ts
@@ -1,0 +1,1 @@
+export const prerender = true;
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -11,16 +11,19 @@
 				"@amplitude/analytics-browser": "^2.3.1",
 				"@auth/core": "^0.18.0",
 				"@auth/sveltekit": "^0.3.11",
+				"@codemirror/lang-json": "^6.0.1",
 				"@mdi/js": "^7.3.67",
 				"@smui/fab": "^7.0.0-beta.15",
-				"@zeno-ml/zeno-instance-views": "^0.0.8",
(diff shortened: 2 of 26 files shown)


frontend/src/routes/playground/+layout.ts
```diff
@@ -0,0 +1 @@
+export const prerender = true;
```
Sparkier, line 1: I don't think we need this now that we're not using GitHub pages anymore.

Sparkier (commented):
I assume this is just copying over stuff, right? No in depth review needed? 

There seem to be 2 package-lock files.

frontend/src/routes/playground/+layout.ts
cabreraalex: Are you sure? I think sveltekit will still pre-render this if it can and it'll be faster

cabreraalex:
> I assume this is just copying over stuff, right? No in depth review needed?
> 
> There seem to be 2 package-lock files.

Yup didn't change much except the paths and imports, all seems to be working. I'll check the lockfiles.

frontend/src/routes/playground/+layout.ts
Sparkier: Can definitely leave it if we think it's better just done need it anymore I think.

Sparkier (approved):

Sparkier:
Will need to update FAQ and zenoml.com

cabreraalex:
[ZEN-334 : update zenoml.com and FAQ with new instance view links](https://linear.app/zenoml/issue/ZEN-334/update-zenomlcom-and-faq-with-new-instance-view-links)

PR #429, opened 2023-11-13
fix: responsive max table width

Sparkier:
# Description

I think this is a 

fix ZEN-327

can you test this as well @cabreraalex ? All the cases I tried worked.


cabreraalex:
Lol I hate that that's all it took I'm awful at css

PR #431, opened 2023-11-13
feat: redo routing to use slug and id or uuid

cabreraalex:
# Description
Should test thoroughly locally and on dev before deploying.

Fix ZEN-339


backend/zeno_backend/database/select.py
```diff
@@ -815,7 +835,7 @@ def project_from_uuid(project_uuid: str) -> Project | None:
             [project_uuid],
         )
         if len(project_result) == 0:
-            raise Exception("Project does not exist.")
+            return None
```
Sparkier, line 838: Exceptions seem like a much better way to communicate the cause of the error compared to just returning None.

backend/zeno_backend/server.py
```diff
@@ -404,18 +386,29 @@ def get_project_state(
...
     @api_app.get(
-        "/report/{owner}/{report}", response_model=ReportResponse, tags=["zeno"]
+        "/report-name/{owner_name}/{report_name}",
+        response_model=ReportResponse,
+        tags=["zeno"],
     )
-    def get_report(owner_name: str, report_name: str, request: Request):
-        rep = select.report(owner_name, report_name, util.get_user_from_token(request))
+    def get_report_by_name(owner_name: str, report_name: str, request: Request):
+        report_id = select.report_id(owner_name, parse.unquote(report_name))
```
Sparkier, line 394: OOC: do we need this parse? Asking since we did not have it before. Also, would it make sense to add for owner_name as well?

Sparkier (approved):
Do we want the user name in the URL as well? Probably not relevant, right?

backend/zeno_backend/database/select.py
cabreraalex: All of the calling functions expect it to return None. Maybe can do another PR to fix all these functions taht return none?

backend/zeno_backend/server.py
cabreraalex: I think we do... can update the other ones.

PR #432, opened 2023-11-13
fix: support heatmap

cabreraalex:
# Description

Fix ZEN-209

### DIFF
--- a/backend/zeno_backend/processing/filtering.py
+++ b/backend/zeno_backend/processing/filtering.py
@@ -115,6 +115,8 @@ def table_filter(
     filter_result: sql.Composed | None = None
     if filter_predicates is not None and len(filter_predicates.predicates) > 0:
         filter_result = filter_to_sql(filter_predicates, project, model)
+        if filter_result == sql.Composed([]):
+            filter_result = None
 
     if data_ids is not None and len(data_ids) > 0:
         db = Database()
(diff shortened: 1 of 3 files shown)


backend/zeno_backend/processing/filtering.py
```diff
@@ -115,6 +115,8 @@ def table_filter(
     filter_result: sql.Composed | None = None
     if filter_predicates is not None and len(filter_predicates.predicates) > 0:
         filter_result = filter_to_sql(filter_predicates, project, model)
+        if filter_result == sql.Composed([]):
+            filter_result = None
```
Sparkier, line 119: Should we do this in `filter_to_sql`? Don't know if there are more places where this is needed.

frontend/src/lib/components/chart/chart-page/encoding/chart-encoding/encoding-components/ModelsEncodingDropdown.svelte
```diff
@@ -26,12 +25,10 @@
...
-		renderer={svelecteRenderer}
-	/>
-</EncodingContainer>
+<Svelecte
+	style="width: 280px; flex:none;"
+	bind:value
+	{options}
+	on:change={valueSelected}
+	renderer={svelecteRenderer}
+/>
```
Sparkier, line 34: Should we add a `container` variable here similar to how we do it for [Metrics](https://github.com/zeno-ml/zeno-hub/blob/e465352c5cbc3d87c1f615cc25d1d2d1adfd2a01/frontend/src/lib/components/chart/chart-page/encoding/chart-encoding/encoding-components/MetricsEncodingDropdown.svelte)?

Sparkier (changes requested):
Could you explain why this is more of a bandaid fix and what is the underlying problem? Somewhat hard to review otherwise.

backend/zeno_backend/processing/filtering.py
cabreraalex: Yup good point, figured out how to do it in the chart.

PR #437, opened 2023-11-15
fix: crazy report behavior with requests

cabreraalex:
# Description
Fixes every element re-rendering in reports on edit.


Sparkier (approved):
WOW :D

PR #438, opened 2023-11-15
fix: replace returns with throws

cabreraalex:
# Description
Recommended way to do it: https://fastapi.tiangolo.com/tutorial/handling-errors/

### DIFF
--- a/backend/zeno_backend/routers/account.py
+++ b/backend/zeno_backend/routers/account.py
@@ -4,7 +4,6 @@ from fastapi import (
     APIRouter,
     Depends,
     HTTPException,
-    Response,
     status,
 )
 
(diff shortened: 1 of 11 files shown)


backend/zeno_backend/routers/account.py
```diff
@@ -117,10 +116,16 @@ def get_user_organizations(current_user=Depends(util.auth.claim())):
 
     Returns:
         list[Organization]: all organizations the user is a member of.
+
+    Raises:
+        HTTPException: error if the user is not found.
```
Sparkier, line 121: Nit: we normally did raises before returns.

Sparkier (approved):
Awesome! Already did this for some when separating the files but not systematically.

PR #442, opened 2023-11-16
feat: add matching secondary color

Sparkier:
# Description

To me, the teal color does not work well with the purple. Could just be me but I think color theory supports my thinking.

The teal is not [complimentary](https://www.sessions.edu/color-calculator-results/?colors=6a1b9a,9a831b) or [related in any other way to the purple](https://www.sessions.edu/color-calculator-results/?colors=6a1b9a,9a1b8b,2a1b9a).

Since we're talking about something like a secondary color, I think an analogous harmony would work well. I picked the [blue here](https://www.sessions.edu/color-calculator-results/?colors=6a1b9a,9a1b8b,2a1b9a) for now which I think looks good.

Happy to discuss further, though.

### DIFF
--- a/frontend/src/lib/components/metadata/SelectionBar.svelte
+++ b/frontend/src/lib/components/metadata/SelectionBar.svelte
@@ -52,7 +52,7 @@
 				{:else}
 					<div class="flex items-center" style="margin-right: 10px">
 						<p style="margin: auto; margin-right: 10px">Editing</p>
-						<div class="my width-fit margin-auto mx-1 rounded bg-greenish-light px-2.5 py-1">
+						<div class="my width-fit margin-auto mx-1 rounded bg-secondary-light px-2.5 py-1">
 							{$editTag.tagName}
 						</div>
 					</div>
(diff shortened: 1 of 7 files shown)


cabreraalex:
Agreed on color, but this looks a bit too similar to purple, I barely noticed they were different?

cabreraalex:
Screenshot 2023-11-16 at 08 40 55

Sparkier:
Yeah they're not super different esp. the light versions we use there...

Sparkier:
how did you get to this? I always paste the existing purple to get another color that might work from these tools. Don't think the purple you have there is used by us, right?

cabreraalex:
The color? I just googled complements to purple and it looked nice lol

Sparkier:
Ahh ok I think we should use the purple we're actually using not just any purple probably.

I've been pasting color values here: https://www.sessions.edu/color-calculator/

cabreraalex:
What about something like #e4f3ff it looks nice to me. And maybe we only have the light color for highlighting tags in the histograms and tabs, but don't do a dark version for the done button, that's just purple.

Sparkier:
kinda like it:
image

PR #443, opened 2023-11-16
feat: add a help button throughout many parts of hub

Sparkier:
# Description

fix ZEN-343

https://github.com/zeno-ml/zeno-hub/assets/5690524/41823a49-29bd-4bea-b14b-58cbdc24bbcf

### DIFF
--- a/frontend/src/lib/components/general/Help.svelte
+++ b/frontend/src/lib/components/general/Help.svelte
@@ -1,0 +1,81 @@
+<script lang="ts">
+	import { browser } from '$app/environment';
+	import { clickOutside } from '$lib/util/clickOutside';
+	import { mdiHelp } from '@mdi/js';
+	import IconButton, { Icon } from '@smui/icon-button';
+	import Paper, { Content } from '@smui/paper';
+	import { fade } from 'svelte/transition';
+
+	export let docsLink = 'https://zenoml.com/docs/intro';
+	export let optionsBelow = false;
+
+	let showOptions = false;
+</script>
+
+<div
+	class="relative z-20 flex flex-col items-end rounded-full border border-primary-dark"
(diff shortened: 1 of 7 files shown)


cabreraalex:
This is EPIC

frontend/src/lib/components/general/Help.svelte
```diff
@@ -0,0 +1,81 @@
...
+	let showOptions = false;
+</script>
+
+<div
+	class="relative z-20 flex flex-col items-end"
+	transition:fade={{ delay: 1000, duration: 1000 }}
+>
+	<IconButton
+		class="border border-primary-light opacity-25 transition hover:opacity-100"
+		on:click={() => (showOptions = !showOptions)}
```
cabreraalex, line 21:
```suggestion
		class="border-primary-lighter rounded-full border bg-primary-mid opacity-50 transition hover:opacity-100"
```

cabreraalex:
I like it. Two opinions:
* Can we make it a bit darker? Really hard to find right now if you don't know where to look. Left a suggestion
* Another option is putting it next to the account button? Would add less clutter and easier to find? but not as distinctive. can go either way.

cabreraalex:
If we leave it on the bottom right, we should have the margin line up with the header on the home page and reports (especially x axis) it's bothering me a bit lol

Sparkier:
image

Sparkier:
I think I like this even better:
image

Sparkier:
I prefer the "always tucked to the right edge of the browser one.

I tried next to the account button already yesterday as well but looked super ugly.

cabreraalex:
Love the circle one

cabreraalex (approved):
LGTM - maybe we add it to the account page?

PR #444, opened 2023-11-16
fix: don't delete old metric if a new one is added

Sparkier:
# Description

fix ZEN-344

### DIFF
--- a/backend/zeno_backend/database/update.py
+++ b/backend/zeno_backend/database/update.py
@@ -6,6 +6,7 @@ from psycopg import sql
 from zeno_backend.classes.chart import Chart, ParametersEncoder
 from zeno_backend.classes.filter import PredicatesEncoder
 from zeno_backend.classes.folder import Folder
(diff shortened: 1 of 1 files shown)


backend/zeno_backend/database/update.py
```diff
@@ -226,11 +227,47 @@ def project_metrics(project_config: Project):
...
+                    columns=metric[3],
+                ),
+                metric_results,
+            )
+        )
+        new_metrics = project_config.metrics
+
+        for metric in metrics:
+            # check if the metric matches one to be added
+            index = next(
```
cabreraalex, line 249: nit: this is a bit unreadable, I don't really know what it's doing. Is there a simpler way to do this logic? If not and it works now worries.

Sparkier: Hmm could also loop but less pretty I'd say....

It's just getting the next element that matches the condition. If not element matches the condition, returns None.

cabreraalex (approved):

PR #446, opened 2023-11-17
style: make home cards match primary and secondary colors

Sparkier:
# Description

Could also use these colors on home. Think it looks pretty good compared to the grey.

image

image


cabreraalex:
It's a little easter colored but could be convinced. Maybe we tap in the designers to help test variations of the secondary color.

Sparkier:
Updated with the colors Kathy proposed. I'm not sure about it...

image

Sparkier:
Still think I like a lighter blue better:
image

cabreraalex:
Screenshot 2023-12-08 at 12 25 53

cabreraalex (approved):

PR #452, opened 2023-11-18
fix: flaky report edits, remove dnd

cabreraalex:
# Description
Drag to move elements was buggy and slow, not even useful since we didn't auto scroll. Removed it and fixed binding to keep frontend and backend synced.

Fix ZEN-351

### DIFF
--- a/frontend/src/routes/(app)/report/[id]/[name]/+page.svelte
+++ b/frontend/src/routes/(app)/report/[id]/[name]/+page.svelte
@@ -30,7 +30,7 @@
 		zenoClient
 			.addReportElement(data.report.id, {
 				type: ReportElementType.TEXT,
-				data: 'new element I',
+				data: 'new element',
 				position: elementIndex
 			})
 			.then((res) => {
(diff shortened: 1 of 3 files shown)


frontend/src/routes/(app)/report/[id]/[name]/+page.svelte
```diff
@@ -32,15 +30,15 @@
 		zenoClient
 			.addReportElement(data.report.id, {
 				type: ReportElementType.TEXT,
-				data: 'new element',
```
Comment (Sparkier) [inline frontend/src/routes/(app)/report/[id]/[name]/+page.svelte:35]: Nit: if we do this numbered thing and don't count up, what's the point?

Sparkier (changes requested):

frontend/src/routes/(app)/report/[id]/[name]/+page.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/report/[id]/[name]/+page.svelte]: Oops that was an accident was debugging.

Sparkier (approved):

PR #455, opened 2023-11-18
fix: various tag fixes

cabreraalex:
# Description

* Move selection to top chips bar, change color
* Fix empty slice updates
* Don't show checkboxes when logged out
* Remove extra options space on tag and slice bar when logged out

### DIFF
--- a/frontend/src/lib/components/instances/Instances.svelte
+++ b/frontend/src/lib/components/instances/Instances.svelte
@@ -29,7 +29,7 @@
 	const zenoClient = getContext('zenoClient') as ZenoService;
 
 	let selected = 'list';
-	let currentResult: Promise<GroupMetric[] | undefined> = new Promise(() => undefined);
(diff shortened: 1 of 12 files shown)


Sparkier (changes requested):
High level comments:

I find the position confusing tbh. If it's up there I expect it to filter, which it does not by default.
image

We currently treat this as an or filter with the other filter. Might be the right move but makes it more confusing for me that it's up there (same for plain tags btw).
image

Maybe we should have two hierarchies here generally? Tag filters and are generally or-connected while slice-based filters are and-connected.

Sparkier:
Maybe we can separate the fixes in this PR from the style changes? Might take a bit longer to get the style right...

cabreraalex:
happy to do either - maybe let's discuss Monday morning

Sparkier (approved):

PR #468, opened 2023-11-20
feat: show instance id

Sparkier:


cabreraalex:
Hmm playing around with it I can go either way on including it. I think it wastes a good amount of space? but could be useful... 

If we include it we should make it a bit darker (very very hard to see now) and maybe vertical align with the checkbox for selecting?

Sparkier:
Removed a bit of the spacing and made it align with the checkmark. I think it's ok space-wise otherwise the question is do we want it at all which now I'm in favor of.

I don't think we should change the color tbh. It should be very subtle and the other option would be `text-grey-dark` which I think is much too similar to the main text color:

image

cabreraalex (approved):

PR #469, opened 2023-11-21
build: github actions testing

cabreraalex:
# Description

Run github actions with docker compose to run playwright tests.

Fix ZEN-189


.github/workflows/ci.yml
```diff
@@ -42,12 +42,11 @@ jobs:
     steps:
       - uses: jakebailey/pyright-action@v1
 
-  test:
```
Sparkier, line 45: should we move this to the backend part? It's not really testing anything just installing. Then there's the question if we need this stuff at all (as well for the frontend) since our integration tests should fail if the install fails. Can likely save some compute here without loosing test coverage.

Sparkier (changes requested):

.github/workflows/ci.yml
cabreraalex: sgtm, we probably want python unit tests at some point but can add that back later.

Sparkier (changes requested):
MVP!

Nit: Seems like we have the test data sql file twice and one of them is ~27 mb. Can we remove one?

Could also try to get it smaller but not super important IMO.

cabreraalex:
> MVP!
> 
> Nit: Seems like we have the test data sql file twice and one of them is ~27 mb. Can we remove one?
> 
> Could also try to get it smaller but not super important IMO.

good catch, frontend one was a mistake. It's big b/c it has the whole translation project, but I think good to test with a large, real-world examle.

Sparkier (approved):

PR #474, opened 2023-11-21
feat: add infinite scrolling to home and explore

Sparkier:
# Description

Loads 20 cards immediately and 20 more whenever the last card becomes visible. Could probably be optimized further but I think it feels pretty good already.

fix ZEN-356


cabreraalex:
Looks gooood. Some thoughts:

* I feel like we should still do the initial fetch in `page.ts`, that way you don't get a blank page first and have to wait for it to load. This is also better for SEO, which is minor but could be a factor.
* Relatedly, before it loads we show "no projects" which is very confusing.
* Can we add padding to the bottom of the page so the bottom of the last cards isn't touching the bottom?

Sparkier:
Fixed. This padding bottom thing has been bothering me as well.

cabreraalex:
I think we can get rid of `updateEntries` and its call in both files now? It calls homeentriest way too many times

cabreraalex (changes requested):
left changes

Sparkier:
> I think we can get rid of `updateEntries` and its call in both files now? It calls homeentriest way too many times 

We need it for filter and sort changes, no?

cabreraalex:
oh lol duh. idk if we can prevent the extra calls at the start, otherwise fine

cabreraalex (approved):

PR #475, opened 2023-11-22
fix: allow setting only valid regexes

Sparkier:


cabreraalex:
this might be too limiting because the regex check uses a different regex syntax than the one used by the database

Sparkier:
I think I might have crashed the frontend with invalid ones.

cabreraalex (approved):

cabreraalex:
Should maybe also limit it in the slice creation menu too then

cabreraalex (approved):

PR #483, opened 2023-11-26
fix: ensure consistnet order to prevent duplicate keys

Sparkier:
# Description

This could break the frontend as we had duplicate keys when the order was not deterministic.


cabreraalex:
Should we also update recent sorting to secondary sort by likes?

PR #485, opened 2023-11-26
fix: make instances always reflect current model

Sparkier:
Instances sometimes did not update when changing the model. This has been fixed by adding an identifier for the model column to the list. Additionally, we don't need to reload the data when changing the model. The table stays the same.


cabreraalex (approved):

cabreraalex (approved):

cabreraalex:
This works but I feel like it shouldn't? Why does it still update the table with the right model when you change it now?

Sparkier:
Do you mean because we removed the reactive table reload on model change? That was never necessary. We've always been loading the whole table with all columns for all models.

PR #495, opened 2023-11-27
fix: fix backend warnings

Sparkier:


cabreraalex:
This looks like two separate PRs?

Sparkier:
Hmm it's all mainly to silence warnings when running the backend.

Sparkier:
What would you separate out?

cabreraalex:
The system-name stuff from the --factory and removing the run server stuff but no worries it's fine

cabreraalex:
integration tests might actually be doing something?

cabreraalex:
Did some debugging - this breaks because slices are stored with "modelName" in the database so existing project slices fail to render the model.

INTEGRATION TESTS WORK!

cabreraalex:
idk how to fix, or if it's worth the name refactor

PR #497, opened 2023-11-28
fix: fix organizations table in account

cabreraalex:
# Description

Was calling wrong function, also add popup for naming.

Fix ZEN-375

### DIFF
--- a/frontend/src/lib/components/settings/OrganizationsTable.svelte
+++ b/frontend/src/lib/components/settings/OrganizationsTable.svelte
@@ -1,12 +1,16 @@
 <script lang="ts">
+	import { invalidate } from '$app/navigation';
 	import { tooltip } from '$lib/util/tooltip';
 	import type { Organization, User, ZenoService } from '$lib/zenoapi';
 	import { mdiClose, mdiCog, mdiLogout, mdiPlus } from '@mdi/js';
-	import { Icon } from '@smui/button';
+	import Button, { Icon } from '@smui/button';
 	import IconButton from '@smui/icon-button/src/IconButton.svelte';
+	import { Content } from '@smui/paper';
+	import Textfield from '@smui/textfield';
 	import { getContext } from 'svelte';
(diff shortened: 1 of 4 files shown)


frontend/src/routes/(app)/account/+page.svelte
```diff
@@ -30,7 +30,7 @@
 
 <div class="flex w-full flex-col">
 	<Header user={data.user} />
-	<div class="mx-6 mb-6">
+	<div class="mx-6 mb-6 overflow-y-scroll">
```
Sparkier, line 33: this should be auto instead of scroll, no?

frontend/src/lib/components/settings/OrganizationsTable.svelte
```diff
@@ -15,18 +19,55 @@
...
+							invalidate('app:organizations');
+							showNewOrganizationPopup = false;
+							reportName = '';
+						})}
+				disabled={reportName.length === 0}
+			>
+				Create
+			</Button>
+		</Content>
+	</Popup>
```
Sparkier, line 62: Move to a file in the popups folder?

frontend/src/lib/components/settings/OrganizationsTable.svelte
```diff
@@ -15,18 +19,55 @@
 
 	let organizationToEdit: Organization | undefined;
 	let showConfirmDelete = false;
+	let showNewOrganizationPopup = false;
 	let organizationToDelete: Organization | undefined;
+	let reportName = '';
```
Sparkier, line 24: nit: organizationName

frontend/src/lib/components/settings/OrganizationsTable.svelte
Sparkier: Think so, should do them all the same. Also with keydown.

Sparkier (changes requested):

Sparkier (approved):

PR #500, opened 2023-11-29
fix: rename model_ to system_ to silence warning and to unify naming

Sparkier:
# Description

I think we should do this in the DB. Otherwise we'll drag along this warning forever:
image

It is as simple as going to the report_elements table and replacing `model_name` with `system_name` for all elements that have a slice or tag in the `data` column.


Sparkier:
**WE NEED TO UPDATE THE DB CONTENT ON MERGE**

This passes all tests but only because the test data is already updated. Need to rename `modelName` to `systemName` as described above.

Sparkier:
106 in prod. The alternative is we just live with the warning. I think it's worth doing the 100 edits but fine either way.

Sparkier:
ping @cabreraalex, if we want to do it we should do it now and probably at a time we don't have a ton of traffic

cabreraalex:
SGTM if you still have access to the prod database you could do this tomorrow morning your time

PR #503, opened 2023-11-30
refactor: change all DB calls to async

cabreraalex:
# Description

Massive change should be thoroughly tested.


backend/zeno_backend/processing/metrics/map.py
```diff
@@ -24,32 +24,34 @@ def count(
...
         GroupMetric: count of datapoints matching the specified filter.
     """
-    with Database() as db:
-        num_total = db.execute_return(
-            sql.SQL("SELECT COUNT(*) FROM {}").format(sql.Identifier(project))
-            if filter is None
-            else sql.SQL("SELECT COUNT(*) FROM {} WHERE ").format(
-                sql.Identifier(project)
+    async with db_pool.connection() as db:
+        async with db.cursor() as cur:
```
Sparkier, line 28: We're repeating this in every function. Should we move it to a function or something? Maybe not worth would make one line out of two....

backend/zeno_backend/server.py
```diff
@@ -43,6 +44,8 @@ def get_server() -> FastAPI:
     Returns:
         FastAPI: FastAPI endpoint
     """
+    _ = db_pool.open()
+
```
Sparkier, line 48: 1. doesn't this need to be awaited?
2. why do we need the `_ =`
3. Should we move this to a function in the DB file instead that is called when getting the pool?

Sparkier (changes requested):
Generally this seems ready just not sure how we can systematically test as well. I guess one problem is that our setup with dev and prod is not really working since we're not using dev. Otherwise I would have said we merge it and don't deploy for a day or two.

backend/zeno_backend/processing/metrics/map.py
cabreraalex: Yeah thought about this but some also use the connection itself and stuff, idk felt like it's easier just the raw way

backend/zeno_backend/database/copy.py
```diff
@@ -25,165 +25,177 @@ def project_copy(project_uuid: str, copy_spec: ProjectCopy, user: User):
...
+        async with conn.cursor() as cur:
+            # Get project to copy from.
+            await cur.execute(
+                "SELECT uuid, name, view, samples_per_page, public, "
+                "description FROM projects WHERE uuid = %s;",
+                [project_uuid],
+            )
+            project_result = await cur.fetchall()
+            if len(project_result) == 0:
+                raise Exception("Project does not exist.")
```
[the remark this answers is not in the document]
cabreraalex: Sold

backend/zeno_backend/server.py
cabreraalex: Done!

Sparkier (approved):

PR #509, opened 2023-11-30
fix: check whether a user is a editor before making a change

Sparkier:
Previously, we only checked for valid access on load. However, we never checked on update whether someone actually had edit rights.


cabreraalex:
maybe wait for the async change and update this?

Sparkier:
Doesn't really depend right? Will create conflicts, though.

cabreraalex (approved):

PR #513, opened 2023-12-01
fix: add project_uuid to delete calls

Sparkier:
Forgot this in the other PR...


Sparkier:
Awesome, saw you already fixed this @cabreraalex

PR #514, opened 2023-12-01
refactor: greatly simplify test data so it can be edited and looked a…

Sparkier:
…t more easily


DEVELOPMENT.md
```diff
@@ -66,3 +66,15 @@ To do so, you need to have the following environment variables set:
...
+
+We are using playwright to test Zeno full-stack.
+To spin up a dummy database, server, and frontend, there is a separate test `docker compose` file that can be used to run tests.
+You should stop your dev docker containers to prevent port clashes.
+
+You will need to slightly change your `.env` file.
+Use `PUBLIC_BACKEND_ENDPOINT = "http://zeno-backend-test:8000"` for the public backend endpoint, and set the `HUB_USERNAME` and `HUB_PASSWORD` for the test account.
+You can get these variables from someone on the Zeno team.
+
+Then, you can either run playwrite from the frontend using `npm run test` or use the playwright vscode extension which gives you some more debugging options.
```
cabreraalex, line 80: sp: playwrite

DEVELOPMENT.md
```diff
@@ -66,3 +66,15 @@ To do so, you need to have the following environment variables set:
...
 - DOCKER_PASSWORD: The password for the username
+
+## Running Tests
+
+We are using playwright to test Zeno full-stack.
+To spin up a dummy database, server, and frontend, there is a separate test `docker compose` file that can be used to run tests.
+You should stop your dev docker containers to prevent port clashes.
+
+You will need to slightly change your `.env` file.
+Use `PUBLIC_BACKEND_ENDPOINT = "http://zeno-backend-test:8000"` for the public backend endpoint, and set the `HUB_USERNAME` and `HUB_PASSWORD` for the test account.
```
cabreraalex, line 77: Is there a way to avoid this? like have a different .env.test file we read instead?

docker-compose-test.yml
```diff
@@ -1,15 +1,19 @@
...
       context: ./frontend
-      dockerfile: Dockerfile.prod
+      dockerfile: Dockerfile.dev
+    volumes:
+      - ./frontend:/app
+      - /app/node_modules/
     ports:
       - "5173:5173"
-    command: npm run preview -- --port 5173 --host 0.0.0.0
+    command: ["npm", "run", "dev", "--", "--host", "0.0.0.0"]
```
cabreraalex, line 14: Why dev? I had it do preview because it gets rid of the flakyness. It's flaky because of vite doing HMR stuff I think.

frontend/playwright.config.ts
```diff
@@ -1,7 +1,7 @@
 import type { PlaywrightTestConfig } from '@playwright/test';
 import dotenv from 'dotenv';
 
-dotenv.config();
+dotenv.config({ path: './.env' });
```
cabreraalex, line 4: oh nice, if we do this we can do the `.env.test` thing above?

frontend/src/lib/api/util.ts
```diff
@@ -7,6 +7,11 @@ export function getEndpoint() {
 			return '/dockerzeno/api';
 		}
 		return 'http://zeno-backend:8000/api';
+	} else if (env.PUBLIC_BACKEND_ENDPOINT == 'http://zeno-backend-test:8000') {
```
cabreraalex, line 10: oof this is bad... any way around it?

frontend/tests/login.ts
```diff
@@ -0,0 +1,13 @@
+import type { Page } from '@playwright/test';
+
+export async function login(page: Page) {
+	await page.goto('/login');
+	await page.waitForTimeout(1000);
```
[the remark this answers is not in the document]
cabreraalex: see note about vite above

frontend/vite.config.ts
```diff
@@ -9,6 +9,12 @@ export default defineConfig({
 				changeOrigin: true,
 				secure: false,
 				rewrite: (path) => path.replace(/^\/dockerzeno/, '')
+			},
+			'/testdockerzeno': {
```
cabreraalex, line 13: with vite preview shouldn't need this.

frontend/src/lib/api/util.ts
cabreraalex: This shouldn't be needed with vite preview

cabreraalex (changes requested):

DEVELOPMENT.md
Sparkier: Tried for a bit but gave up since we're also using the SvelteKit native functionality of accessing variables in `.env`. Maybe we can find a way but probably for a future PR?

DEVELOPMENT.md
Sparkier: I don't think that's correct.

Sparkier: https://playwright.dev

frontend/playwright.config.ts
Sparkier: I don't think so, see above reasoning.

DEVELOPMENT.md
cabreraalex: This module provides access to runtime environment variables, as defined by the platform you're running on. For example if you're using [adapter-node](https://github.com/sveltejs/kit/tree/master/packages/adapter-node) (or running [vite preview](https://kit.svelte.dev/docs/cli)), this is equivalent to process.env. This module only includes variables that do not begin with [config.kit.env.publicPrefix](https://kit.svelte.dev/docs/configuration#env) and do start with [config.kit.env.privatePrefix](https://kit.svelte.dev/docs/configuration#env) (if configured).

I think it would work with preview if we set the env files variables in the docker file?

Sparkier: Where is this from?

PR #516, opened 2023-12-01
feat: larger instance size option

cabreraalex:
# Description
Add a larger size option for more complex instances

Fix ZEN-321


frontend/src/lib/components/instances/ComparisonView.svelte
```diff
@@ -180,7 +180,7 @@
 </script>
 
 <div class="h-full w-full overflow-auto" bind:this={instanceContainer}>
-	<table class="mt-2 w-full table-fixed">
```
Sparkier, line 183: Why the min-w? Shouldn't this just scale and scroll anyway?

frontend/src/routes/playground/+page.svelte
```diff
@@ -42,9 +43,17 @@
...
+			<img src="/zeno-logo.png" class="h-8" alt="diamond tesselation logo" />
+		</a>
+		<h1 class="ml-3 text-2xl font-bold">Instance View Playground</h1>
+		<div class="ml-auto mr-2">
+			<Button variant="raised" href="https://zenoml.com/docs/views/">Read the Docs</Button>
+		</div>
+	</div>
 	<div class="flex w-full items-center pb-4">
-		<h2 class="pr-4 text-xl">Examples:</h2>
+		<h2 class="pr-4 text-xl">Standard Views:</h2>
```
Sparkier, line 56: Separate PR? Fine for now but generally would.

frontend/src/lib/instance-views/schema.ts
```diff
@@ -71,4 +71,5 @@ export interface ViewSchema {
 	label?: ViewUnion;
 	output?: ViewUnion;
 	displayType?: DisplayType;
+	size?: 'medium' | 'large';
```
Sparkier, line 74: nit: do we want to write it out or just do `s`, `m`, `l`?

frontend/src/lib/instance-views/InstanceView.svelte
```diff
@@ -67,21 +67,23 @@
 		tabindex="0"
 		role="button"
 	>
-		<div class="flex h-10 w-full items-center justify-between pl-4">
-			<div class="text-xs text-grey-light">
-				{entryId}
+		{#if entryId}
```
Sparkier, line 70: We should change entryId to fall back to undefined probably. Otherwise pretty opaque.

frontend/src/routes/playground/+page.svelte
```diff
@@ -42,9 +43,17 @@
 </div>
 
 <div class="flex h-full w-full flex-col p-4">
-	<h1 class="mb-2 text-2xl font-bold">Zeno View Playground</h1>
+	<div class="mb-4 flex h-12 items-center">
+		<a href="/" class="shrink-0">
+			<img src="/zeno-logo.png" class="h-8" alt="diamond tesselation logo" />
+		</a>
+		<h1 class="ml-3 text-2xl font-bold">Instance View Playground</h1>
```
Sparkier, line 50: This looks weird now. Maybe move below the header or remove?

frontend/tailwind.config.js
```diff
@@ -41,6 +41,10 @@ export default {
 		extend: {
 			gridTemplateColumns: {
 				home: 'repeat(auto-fill, minmax(275px, 1fr))'
+			},
+			minWidth: {
+				'small-instance': '400px',
+				'large-instance': '800px'
```
Sparkier, line 47: Would move this to the instance view directly and remove from here.

frontend/src/lib/components/instances/TableView.svelte
```diff
@@ -175,7 +175,7 @@
 							{#if instanceHidden}
 								<p class="text-center">...</p>
 							{:else}
-								<div class="instance">
+								<div class="min-w-small-instance">
```
Sparkier, line 178: Would remove the div and move the sizing to the instance view itself.

frontend/tailwind.config.js
Sparkier: Only used once anyway.

frontend/src/lib/components/report/elements/SliceElement.svelte
```diff
@@ -17,20 +18,23 @@
...
 
 	const zenoClient = getContext('zenoClient') as ZenoService;
+
 	let sliceElementSpec: SliceElementSpec | undefined;
 	let sliceElementOptions: SliceElementOptions | undefined;
 	let table: Record<string, string | number | boolean>[] | undefined = [];
 	let page = 0;
 
 	$: updateSliceElementSpec(element.data as string);
+	$: viewSchema = JSON.parse(sliceElementOptions?.project.view || '{}') as ViewSchema;
```
Sparkier, line 28: This will fail if the JSON.parse fails because it can't be cast. Will probably crash frontend then.

frontend/src/lib/components/instances/ComparisonView.svelte
cabreraalex: The instance views shrink to be kind of tiny in some cases. This forces them to at least fill most of the space.

frontend/src/lib/components/instances/TableView.svelte
cabreraalex: The width in the instance view changes though depending on what view it's in, e.g. reports

frontend/src/lib/instance-views/InstanceView.svelte
cabreraalex: not sure what you mean

frontend/src/routes/playground/+page.svelte
cabreraalex: Yeah probably should have been

frontend/src/routes/playground/+page.svelte
cabreraalex: Hmm looks okay to me? which part is off?

frontend/src/lib/components/report/elements/SliceElement.svelte
cabreraalex: Won't crash b/c it's typescript, but maybe cleaner way to do this?

frontend/src/lib/instance-views/InstanceView.svelte
Sparkier: I think you understood you changed to null.

frontend/src/lib/components/instances/ComparisonView.svelte
Sparkier: Should we rather set a min-w on the instance view? This will not really solve the problem, right? If the table really is less than 1000 it's kinda weird and if there are tons of columns the instance will still shrink to tiny.

Sparkier (changes requested):

frontend/src/lib/components/instances/ComparisonView.svelte
cabreraalex: Hmm don't think I understand. Some instance views can be tiny, do we want to make them bigger than they have to be?

Sparkier (approved):

PR #517, opened 2023-12-01
docs: self-host

cabreraalex:
# Description
Small blurb to describe self-hosting.

Fix ZEN-385


Sparkier (approved):
Btw. I don't think I get the emails from hello@, right? Might be good to add me to that somehow.

cabreraalex:
checked and don't think I can do multiple forwards unfortunately.

afennelly-mitre:
Why do we need to send an email to get the environment variables? Can you not just add those to the description? I cannot use ZenoHub for my use case and **need** to self-host. It would be much faster to experiment with Zeno if the docs were all-inclusive...

PR #518, opened 2023-12-01
feat: cache chart data

cabreraalex:
# Description
Saves chart data as a column for fast loading. Invalidates it when slices, systems, or metrics change.

We could do more fine-grained invalidation but this is already orders of magnitude better, not worth complexity now IMO.

Alter DB as follows:
`ALTER TABLE charts ADD data jsonb;`

Fix ZEN-354

Blocked by #514


backend/zeno_backend/database/select.py
```diff
@@ -1654,14 +1659,17 @@ async def chart(project_uuid: str, chart_id: int) -> Chart:
...
+            detail="Chart could not be found.",
         )
+
     return Chart(
         id=chart_result[0][0],
         name=chart_result[0][1],
         type=chart_result[0][2],
         project_uuid=project_uuid,
         parameters=json.loads(chart_result[0][3]),
+        data=json.dumps(chart_result[0][4]) if chart_result[0][4] is not None else None,
```
Sparkier, line 1672: Don't need the check since we're assigning `None` on fail anyway.

backend/zeno_backend/routers/chart.py
```diff
@@ -41,7 +41,7 @@ async def get_charts(project_uuid: str, request: Request):
 
 @router.get(
     "/chart/{owner}/{project}/{chart_id}",
```
Sparkier, line 43: Old but don't think we need the owner param here.

backend/zeno_backend/database/delete.py
```diff
@@ -119,6 +119,23 @@ async def chart(chart_id: int):
...
+    """
+    async with db_pool.connection() as conn:
+        async with conn.cursor() as cur:
+            await cur.execute(
+                "UPDATE charts SET data = NULL WHERE project_uuid = %s;",
+                [
+                    project_uuid,
+                ],
+            )
+            await conn.commit()
```
Sparkier, line 136: Should we use update here instead? We typically only use delete if we delete entries from the database iirc.

backend/zeno_backend/routers/slice.py
```diff
@@ -215,6 +217,7 @@ async def update_slice(slice: Slice, project: str, request: Request):
         request (Request): http request to get user information from.
     """
     await util.project_editor(project, request)
+    await delete.chart_data(project)
```
Sparkier, line 220: Don't we need to do this for delete slice as well?

docker-compose.yml
```diff
@@ -43,6 +43,7 @@ services:
       POSTGRES_PASSWORD: mypassword
     volumes:
       - ./backend/test-data.sql:/docker-entrypoint-initdb.d/test-data.sql
+    # command: "-c log_statement=all"
```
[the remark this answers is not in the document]
Sparkier: If so, can we add a comment describing what this does and how to use it?

frontend/src/routes/(app)/project/[uuid]/[name]/chart/[chartIndex=integer]/+page.svelte
```diff
@@ -17,8 +17,7 @@
 
 	let isChartEdit: boolean | undefined;
 	let chart = data.chart;
-	let chartData: { table: Record<string, unknown> } | undefined = data.chartData;
-
+	let chartData = JSON.parse(chart.data || '{}');
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[uuid]/[name]/chart/[chartIndex=integer]/+page.svelte:20]: Should this be undefined instead? Probably easier to handle and check for.

frontend/src/routes/(app)/project/[uuid]/[name]/chart/[chartIndex=integer]/+page.svelte
```diff
@@ -61,11 +58,9 @@
...
-			<ChartContainer chartName={chart.name} loading={updatingData}>
-				<svelte:component this={chartMap[chart.type]} {chart} data={chartData} width={900} />
-			</ChartContainer>
-		</div>
-	{/if}
+	<div class={`flex h-full flex-col overflow-auto pl-2`}>
+		<ChartContainer chartName={chart.name} loading={updatingData}>
+			<svelte:component this={chartMap[chart.type]} {chart} data={chartData} width={900} />
+		</ChartContainer>
+	</div>
```
Comment (Sparkier) [inline frontend/src/routes/(app)/project/[uuid]/[name]/chart/[chartIndex=integer]/+page.svelte:65]: If we remove this check, charts can be loaded with empty data, right? Does that make sense?

frontend/src/routes/(app)/report/[id]/[name]/+layout.server.ts
```diff
@@ -32,6 +34,11 @@ export async function load({ cookies, params, url }) {
 			);
 		}
 	}
+	const [charts, slices, tags] = await Promise.all([
+		zenoClient.getChartsForProjects(reportResponse.report.linkedProjects),
+		zenoClient.getSlicesForProjects(reportResponse.report.linkedProjects),
+		zenoClient.getTagsForProjects(reportResponse.report.linkedProjects)
+	]);
```
Comment (Sparkier) [inline frontend/src/routes/(app)/report/[id]/[name]/+layout.server.ts:41]: Should we do this here? Only really relevant if in edit mode, right?

Sparkier (changes requested):

frontend/src/routes/(app)/project/[uuid]/[name]/chart/[chartIndex=integer]/+page.svelte
Comment (cabreraalex) [reply frontend/src/routes/(app)/project/[uuid]/[name]/chart/[chartIndex=integer]/+page.svelte:65]: Added in case.

frontend/src/routes/(app)/report/[id]/[name]/+layout.server.ts
Comment (cabreraalex) [reply frontend/src/routes/(app)/report/[id]/[name]/+layout.server.ts:41]: Not necessarily, we get the chart from this array when rendering them at all.

Agree it's pretty inefficient though. I think future PR we should redo how we do this querying generally.

PR #526, opened 2023-12-04
fix: fix copy bug and only show menu when logged in

Sparkier:
# Description

fix ZEN-394


cabreraalex (approved):

cabreraalex:
check CI

Sparkier:
Think only fails because I changed the env variable.

PR #527, opened 2023-12-04
fix: check element id with project before deleting element

adityamittal13:
# Description

fix ZEN-386

check files: chart.py, slice.py, tag.py, folder.py
- rebased on accident so extra files from today's commits were added


cabreraalex:
you should be able to `git merge origin/main` to get rid of the extra changes

adityamittal13:
just made that change - check now

backend/zeno_backend/routers/chart.py
```diff
@@ -170,4 +170,11 @@ async def delete_chart(project_uuid: str, chart_id: int, request: Request):
...
     """
     await util.project_editor(project_uuid, request)
-    await delete.chart(chart_id)
+    chart = await select.chart(project_uuid, chart_id)
+    if chart is not None:
+        await delete.chart(chart_id)
+    else:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Unauthorized",
```
cabreraalex, line 179: I don't think we need the `detail` message here since the status code represents it, can remove in other files too.

cabreraalex (changes requested):
Looks good! left one comment.

Can we also check this for the `update` calls? e.g. you shouldn't be able to update an element either

adityamittal13:
Just fixed the comment, check now.

backend/zeno_backend/routers/chart.py
```diff
@@ -155,7 +155,10 @@ async def update_chart(chart: Chart, project_uuid: str, request: Request):
         request (Request): http request to get user information from.
     """
     await util.project_editor(project_uuid, request)
-    return await update.chart(chart, project_uuid)
+    if chart.project_uuid == project_uuid:
```
cabreraalex, line 158: I *think* you still have to call `select.chart(chart.id)` and then compare that result with the project_uuid. Someone could make a request with a Chart object that has a project_uuid that is different from the charts'

cabreraalex (changes requested):

backend/zeno_backend/routers/chart.py
adityamittal13: The current select.chart takes in two parameters: the project_uuid and the chart_id. Should I implement a select.chart_by_id similar to how select.slice_by_id works? In it, the SQL query could then just take in the chart id and we'd select for the project_uuid.

cabreraalex: actually you can change `select.chart` to just take the `id`, it doesn't need the UUID. Good catch!

cabreraalex: we should probably change `slice_by_id` to just be `slice` too

adityamittal13: Ok let me do that right now!

adityamittal13:
Should be good now - had to change a couple of areas in the code that used the functions before, but ran locally and everything works.

cabreraalex:
Fantastic, thanks so much! Will approve and merge once I fix the CI

cabreraalex (approved):

Sparkier (changes requested):
Can we add a description to the `HTTPException`s?

Something like "this chart does not belong to your project"

cabreraalex:
aha I can do that for this PR

Sparkier (approved):

PR #530, opened 2023-12-05
build: tests for uploading from client

cabreraalex:
# Description

Runs after playwright, uses the client to upload project, data, systems.

Fixes CI as well.

Blocked by https://github.com/zeno-ml/zeno-client/pull/44


.github/workflows/ci.yml
```diff
@@ -44,18 +44,29 @@ jobs:
...
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
-      - name: Set up secrets
-        run: echo "${{ secrets.ZENO_HUB_ENV }}" > frontend/.env
+      - uses: actions/setup-python@v2
+      - name: Install Zeno Client
+        run: |
+          python -m pip install --upgrade pip
+          pip install zeno-client
+          pip install tqdm
```
cabreraalex, line 61: remove after client update

db/test-data.sql
```diff
@@ -170,3 +171,4 @@ COPY public.report_project (id, report_id, project_uuid) FROM stdin;
 COPY public.user_project (id, user_id, project_uuid, editor) FROM stdin;
 1	1	f93ccc54-f701-4da8-b816-696e166f3274	t
 \.
+SELECT setval('public.user_project_id_seq', COALESCE((SELECT MAX(id) FROM public.user_project), 1));
```
Sparkier, line 174: If we're doing this, should we remove the id column from the copy command? Don't know if this is easier to understand/manipulate, though.

Sparkier (changes requested):

db/test-data.sql
cabreraalex: sure, doesn't matter either way, have to increment still.

Sparkier (approved):

PR #532, opened 2023-12-07
fix: allow duplicate slice names

zaidsheikh:
# Description

Allow duplicate slice names, mainly so we can easily "Create slices for all values" without worrying about existing slice names. This was decided during a discussion on [ZEN-282](https://linear.app/zenoml/issue/ZEN-282/create-chart-from-metadata-histogram). 

(Note: we probably need a way to disambiguate slices with the same name in the chart editor … maybe by including the folder name as well)

# References

fix ZEN-399

### DIFF
--- a/backend/zeno_backend/database/insert.py
+++ b/backend/zeno_backend/database/insert.py
@@ -640,25 +640,18 @@ async def all_slices_for_column(
                     folder_id=None if folder_id is None else folder_id[0],
                 )
                 await cur.execute(
-                    sql.SQL(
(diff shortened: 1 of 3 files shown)


Sparkier:
Hi @zaidsheikh is this ready for review?

zaidsheikh:
I need to push one more commit to fix an issue with charts (they currently assume slice names are unique, so only one slice is shown even if multiple slices with the same name are selected).

Sparkier:
Ah ok sounds good, thank you!

PR #536, opened 2023-12-07
fix: show vstack keys in order specified

cabreraalex:
# Description

Fix ZEN-398

### DIFF
--- a/frontend/src/lib/instance-views/elements/VStack.svelte
+++ b/frontend/src/lib/instance-views/elements/VStack.svelte
@@ -9,7 +9,6 @@
 	let jsonData: Record<string, View> | undefined = undefined;
 	let errorMessage: string | undefined = undefined;
 
-	$: keys = jsonData ? Object.keys(jsonData) : [];
 	$: try {
 		jsonData = JSON.parse(data);
 		errorMessage = undefined;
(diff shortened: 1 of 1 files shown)


frontend/src/lib/instance-views/elements/VStack.svelte
```diff
@@ -27,13 +26,15 @@
...
-			{#if spec.keys && spec.keys[key] !== undefined && spec.keys[key].type !== undefined}
+		{#each Object.keys(spec.keys) as key}
+			{#if key in jsonData}
 				<svelte:component
 					this={resolveElementType(key)}
 					spec={spec.keys[key]}
 					data={typeof jsonData[key] === 'object' ? JSON.stringify(jsonData[key]) : jsonData[key]}
 				/>
+			{:else}
+				<p class="text-error">ERROR: no data for key <b>{key}</b></p>
```
Sparkier, line 37: Don't think we should show an error here. This can be intentional and useful. You could have a key that's just sometimes there or for one type of element you have one key for the other type you have another key.

frontend/src/lib/instance-views/elements/VStack.svelte
```diff
@@ -27,13 +26,15 @@
 	<Error type="Incorrect Data" message={errorMessage} />
 {:else}
 	<div class="flex flex-col">
-		{#each keys as key}
-			{#if spec.keys && spec.keys[key] !== undefined && spec.keys[key].type !== undefined}
+		{#each Object.keys(spec.keys) as key}
+			{#if key in jsonData}
```
Sparkier, line 30: I don't think it makes sense to show them in the order of the spec but rather in the order they are in the json data. This makes it more flexible and views can be combined more freely.

Sparkier (changes requested):

PR #538, opened 2023-12-07
fix: detect nonexistent standard views

cabreraalex:
# Description

Fix ZEN-404


backend/zeno_backend/database/select.py
```diff
@@ -183,7 +183,7 @@ async def projects(
             Project(
                 uuid=res[0],
                 name=res[1],
-                view=match_instance_view(res[3]),
+                view="",
```
cabreraalex, line 186: This can break login for existing projects on the homescreen with bad view specs. Don't need view here anyway

PR #545, opened 2023-12-09
feat: copy links in projects and reports

cabreraalex:
# Description
image


Sparkier (approved):
Might be a bit confusing with the link icon. Maybe people think the project/report is linked to something?

Don't really know what would be better, though... [Maybe a share icon](https://materialui.co/icon/ios-share)? Fine either way.

PR #560, opened 2023-12-11
feat: add display name to users

cabreraalex:
# Description
Fix ZEN-255

Display name for reports and projects. Currently only shown on reports, not on home or in projects - I think just username for those to make sure people aren't impersonating on the home screen?

Defaults to username.

Commands to alter table:

```
ALTER TABLE users ADD COLUMN display_name text;
UPDATE users SET display_name = users.name;
ALTER TABLE users ALTER COLUMN display_name SET NOT NULL;
```

### DIFF
--- a/frontend/src/lib/components/settings/Account.svelte
+++ b/frontend/src/lib/components/settings/Account.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { page } from '$app/stores';
-	import type { ZenoService } from '$lib/zenoapi';
+	import type { User, ZenoService } from '$lib/zenoapi';
 	import Button from '@smui/button/src/Button.svelte';
 	import Textfield from '@smui/textfield';
 	import { getContext } from 'svelte';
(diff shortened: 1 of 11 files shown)


frontend/src/lib/components/settings/Account.svelte
```diff
@@ -1,19 +1,22 @@
...
 	import { page } from '$app/stores';
-	import type { ZenoService } from '$lib/zenoapi';
+	import type { User, ZenoService } from '$lib/zenoapi';
 	import Button from '@smui/button/src/Button.svelte';
 	import Textfield from '@smui/textfield';
 	import { getContext } from 'svelte';
 	import { fade } from 'svelte/transition';
 	import FeatureFlagsPopup from '../popups/FeatureFlagsPopup.svelte';
 
 	export let email: string;
```
Sparkier, line 10: Don't technically need this extra export, right? Can get it from the user.

Sparkier (approved):
One nit and not sure about where to use usernames vs. display names. I feel like display names everywhere is probably fine? Could impersonate in reports, too, and people probably want other people to see their display name everywhere. Esp. in orgs where there might even be naming conventions or people just use their full names so they can be identified. Don't know.

frontend/src/lib/components/settings/Account.svelte
Sparkier: nvm was wrong

PR #561, opened 2023-12-11
feat: multiple authors on reports

cabreraalex:
# Description

Fix ZEN-256

Blocked by #560 

Edits to database:

``` ```


cabreraalex:
This is slightly complicated on how we implement it, kinda gave up for now.

PR #568, opened 2023-12-14
fix: move help to top bar, move project settings to top bar

cabreraalex:
# Description
Why?
1. Makes settings in a consistent place across all views
2. Makes help a bit more discoverable, same in all views, not hacky depending on what is shown where.

Fix ZEN-411

image


frontend/src/lib/components/general/Header.svelte
```diff
@@ -21,7 +21,7 @@
 	let linkCopied = false;
 </script>
 
-<div class="mx-6 my-5 flex h-8 items-center justify-between">
+<div class="mx-6 mb-4 mt-4 flex h-8 items-center justify-between">
```
Sparkier, line 24: nit: my-4

frontend/src/lib/components/general/UserButton.svelte
```diff
@@ -1,34 +1,105 @@
...
+							showOptions = false;
+						}}
+					>
+						<Icon style="font-size: 20px;" class="material-icons">article</Icon>&nbsp;
+						<span class="whitespace-nowrap text-sm">Explore the Docs</span>
+					</button>
+				</Content>
+			</Paper>
+		</button>
 	{/if}
```
Sparkier, line 104: Should we move this in here? Will make it difficult if we ever want to add it to somewhere else (e.g. playground)

Sparkier (changes requested):
I like the settings moving up. Am ok with help also moving but think it's much less prominent now. Might be fine since we're starting to link to docs from other places (e.g. reports).

If we keep it like this, I think we definitely should add it to the playground as well.

cabreraalex:
Made it a component and added to playground!

frontend/src/lib/components/general/HelpButton.svelte
```diff
@@ -1,29 +1,28 @@
...
+	import { Icon } from '@smui/icon-button';
 	import Paper, { Content } from '@smui/paper';
 	import { fade } from 'svelte/transition';
 
-	export let docsLink = 'https://zenoml.com/docs/intro';
-	export let optionsBelow = false;
+	export let showOptions = false;
 
-	let showOptions = false;
+	let docsLink = 'https://zenoml.com/docs/intro';
```
Sparkier, line 11: Used this as export to deep-link to different parts of the docs previously but probably not necessary anymore.

Sparkier (approved):

PR #570, opened 2023-12-14
feat: show a preview of the charts in the chart overview

Sparkier:
# Description

image

fix ZEN-391


Sparkier:
Don't understand the linter errors, I don't get them locally.

Sparkier:
ping @cabreraalex

cabreraalex:
Honestly think it looks pretty rough with the huge fonts, crunched up legends, etc... idk if we actually want this? what do you think?

Sparkier:
There shouldn't be any legends visible in the previews. Don't know if you had the right version.

Sparkier:
Ah the screenshot is outdated btw.

Sparkier:
Actually, I pushed something wrong, but looks like this:

CleanShot 2024-01-11 at 09 57 35

Sparkier:
Could think about removing all the axis labels and text but generally this makes it so much easier to find the chart you're looking for.

Sparkier:
We'll probably have to fix some things once it's deployed but I'd say it's already much better than what we had.

cabreraalex:
Alright yeah it had't updated locally for some reason. It looks good to me, one worry is if you have dozens of charts it could slow down to load, maybe we add infinite scroll or pagination as a todo?

Sparkier:
sgtm, testing speed locally is kinda hard so not sure about it. One thing is that the charts are loaded async anyway I think so that might help.

cabreraalex (approved):

PR #575, opened 2023-12-15
feat: order models, slices, and metrics by name per default

Sparkier:
# Description

Think we should order models, slices, and metrics per default. This is also helpful to group them since similar models also typically have a similar name (e.g. GPT). Also makes it so they're next to each other in charts when all models is selected.

Now it's somewhat random which model appears first.

fix ZEN-413

### DIFF
--- a/backend/zeno_backend/database/select.py
+++ b/backend/zeno_backend/database/select.py
@@ -112,7 +112,8 @@ async def models(project: str) -> list[str]:
         async with conn.cursor() as cur:
             await cur.execute(
                 sql.SQL(
-                    "SELECT DISTINCT model FROM {} WHERE model IS NOT NULL;"
+                    "SELECT DISTINCT model FROM {} WHERE model IS NOT NULL "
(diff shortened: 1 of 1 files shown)


cabreraalex:
Is this not in the order they are uploaded in? or reverse?

For systems I imagine reverse chronological order makes the most sense in most situations, but maybe we can add that later

Sparkier:
> The only reason you usually get a consistent sort order is that the database usually executes the query in the same way.  It's so consistent, in fact, that eventually most developers forget that you can't actually rely on this order. 

So, currently we might get something like this but not necessarily. I also don't know if chronological even makes the most sense, maybe sometimes, but you often want similarly named models next to each other (e.g. first I want to see the GPT model family then the Mistral models, etc.)

cabreraalex:
SGTM! Eventually I think we'll want a more sophisticated model picker with multiple fields you can sort by, search, etc. so maybe a different problem.

cabreraalex (approved):

PR #593, opened 2023-12-19
feat: add chart config

Sparkier:
# Description

towards ZEN-313

Need to run:
```
CREATE TABLE chart_config (
    id integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
    project_uuid text NOT NULL REFERENCES projects(uuid) ON DELETE CASCADE ON UPDATE CASCADE,
    chart_id integer REFERENCES charts(id) ON DELETE CASCADE ON UPDATE CASCADE,
    config jsonb NOT NULL
)
```
to update the DB.

### DIFF
--- a/backend/zeno_backend/database/select.py
+++ b/backend/zeno_backend/database/select.py
@@ -2662,46 +2662,33 @@ async def system_exists(project_uuid: str, system_name: str) -> bool:
         raise HTTPException(status_code=500, detail="Could not check if system exists")
 
 
-async def chart_config(project_uuid: str) -> ChartConfig:
+async def chart_config(
+    project_uuid: str, chart_id: int | None = None
+) -> ChartConfig | None:
     """Get a project's chart configuration.
 
     Args:
         project_uuid (str): the project for which to fetch the config.
+        chart_id (int): the chart for which to fetch the config.
 
     Returns:
         ChartConfig | None: the config if there is one, otherwise None.
     """
     async with db_pool.connection() as conn:
         async with conn.cursor() as cur:
-            await cur.execute(
-                "SELECT config from chart_config WHERE project_uuid = %s;",
-                [project_uuid],
-            )
+            if chart_id is None:
+                await cur.execute(
+                    "SELECT config from chart_config WHERE project_uuid = %s "
+                    "AND chart_id IS NULL;",
+                    [project_uuid],
+                )
+            else:
+                await cur.execute(
+                    "SELECT config from chart_config WHERE chart_id = %s;",
+                    [chart_id],
+                )
             config = await cur.fetchall()
 
     if len(config) == 0:
-        return ChartConfig(project_uuid=project_uuid)
+        return None
     return config[0][0]
-
-
-async def has_chart_config(project_uuid: str) -> bool:
-    """Check if a project has a chart configuration.
-
-    Args:
-        project_uuid (str): the project for which to fetch the config.
-
-    Returns:
-        bool: whether or not the project has a chart config.
-    """
-    async with db_pool.connection() as conn:
-        async with conn.cursor() as cur:
-            await cur.execute(
-                "SELECT id from chart_config WHERE project_uuid = %s;",
-                [project_uuid],
-            )
-            config = await cur.fetchall()
-
-    if len(config) == 0:
-        return False
-
-    return True
--- a/backend/zeno_backend/database/select.py
+++ b/backend/zeno_backend/database/select.py
@@ -6,7 +6,7 @@ from fastapi import HTTPException, status
 from psycopg import sql
 
 from zeno_backend.classes.base import MetadataType, ZenoColumn, ZenoColumnType
-from zeno_backend.classes.chart import Chart
+from zeno_backend.classes.chart import Chart, ChartConfig
 from zeno_backend.classes.filter import FilterPredicateGroup, Join, Operation
 from zeno_backend.classes.folder import Folder
 from zeno_backend.classes.homepage import EntrySort, HomeRequest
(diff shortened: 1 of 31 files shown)


Sparkier:
Not really sure what other properties we want users to be able to configure. Maybe color schemes?

backend/zeno_backend/database/select.py
```diff
@@ -2660,3 +2660,48 @@ async def system_exists(project_uuid: str, system_name: str) -> bool:
...
+        async with conn.cursor() as cur:
+            await cur.execute(
+                "SELECT config from chart_config WHERE project_uuid = %s;",
+                [project_uuid],
+            )
+            config = await cur.fetchall()
+
+    if len(config) == 0:
+        return ChartConfig(project_uuid=project_uuid)
+    return config[0][0]
```
cabreraalex, line 2684: Do we need to cast or anything here? How does Python know this is a ChartConfig? idk types in python are confusing to me

frontend/src/routes/(app)/project/[uuid]/[name]/chart/+page.svelte
```diff
@@ -1,21 +1,27 @@
...
 	import { ChartType, ZenoService } from '$lib/zenoapi';
 	import { mdiPlus } from '@mdi/js';
-	import Button from '@smui/button/src/Button.svelte';
+	import Button from '@smui/button';
 	import { Icon } from '@smui/icon-button';
 	import { getContext } from 'svelte';
 
 	export let data;
 
+	let chartConfigEdit = false;
```
Comment (cabreraalex) [inline frontend/src/routes/(app)/project/[uuid]/[name]/chart/+page.svelte:15]: nit: put under const

cabreraalex (approved):
LGTM! two small comments.

I'm assuming that for the chart-level config when we fetch it we make it "invisible" to the frontend - when loading a chart config we first pull the project config, then pull the chart config and override any set fields"?

cabreraalex:
> Not really sure what other properties we want users to be able to configure. Maybe color schemes?

I think this is fine for now until more requests or our needs come up? I would add chart-level preferences first IMO

backend/zeno_backend/database/select.py
Sparkier: No, don't need to cast.

Sparkier:
Added chart-level prefs that overrule project level.

Sparkier:
Re-requesting review since a lot has changed. Maybe you can test it rq? If not no worries.

frontend/src/lib/components/popups/ChartConfigPopup.svelte
```diff
@@ -0,0 +1,63 @@
...
+			zenoClient.getChartConfig(config.projectUuid, chartId).then((fetchedConfig) => {
+				config = fetchedConfig;
+				dispatch('close');
+			});
+		});
+	}
+
+	function submit(e: KeyboardEvent) {
+		if (e.key === 'Escape') {
+			undoChartConfig();
```
cabreraalex, line 38: can we make this tailwind?

frontend/src/lib/components/popups/ChartConfigPopup.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+	function undoChartConfig() {
+		config = oldConfig;
+		dispatch('close');
+	}
+
+	function resetChartConfig() {
+		zenoClient.deleteChartConfig(config.projectUuid, chartId).then(() => {
+			zenoClient.getChartConfig(config.projectUuid, chartId).then((fetchedConfig) => {
+				config = fetchedConfig;
```
cabreraalex, line 30: same about tailwind. classes should work on here?

frontend/src/lib/components/popups/ChartConfigPopup.svelte
```diff
@@ -0,0 +1,63 @@
...
+		});
+	}
+
+	function undoChartConfig() {
+		config = oldConfig;
+		dispatch('close');
+	}
+
+	function resetChartConfig() {
+		zenoClient.deleteChartConfig(config.projectUuid, chartId).then(() => {
```
cabreraalex, line 28: This needs a higher z-index I think, the vega settings circle is shown on top of it.

cabreraalex (approved):

cabreraalex:
oops sorry VSCode GH left comments in the wrong spot - it's for the HTML part I think should be easy to spot for my comments.

Overall it makes lots of sense and is pretty intuitive. Only thought is maybe we move the "reset" button to the left and make it a different color, name it "Reset to Default" for project-level and "Reset to Project Settings" for chart-level to make it more clear what the reset button does?

frontend/src/lib/components/chart/chart-page/chart-header/SettingsButton.svelte
```diff
@@ -0,0 +1,24 @@
...
+	<ChartConfigPopup
+		bind:config={chartConfig}
+		chartId={chart.id}
+		on:close={() => (chartConfigEdit = false)}
+	/>
+{/if}
+<Button
+	class="ml-auto mr-8 mt-2 shrink-0 sm:mt-0"
+	variant="outlined"
+	on:click={() => (chartConfigEdit = true)}>Settings</Button
```
cabreraalex, line 23: Maybe we call this "Customize Theme" or something? Since we already have the chart-level settings it's a bit confusing.

cabreraalex: or "Chart Configuration" like the title is inside the modal

frontend/src/lib/components/popups/ChartConfigPopup.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+	function submit(e: KeyboardEvent) {
+		if (e.key === 'Escape') {
+			undoChartConfig();
+		}
+	}
+</script>
+
+<svelte:window on:keydown={submit} />
+<Popup on:close>
```
cabreraalex, line 44: This needs a higher z-index, the vega settings overlaps it

frontend/src/lib/components/popups/ChartConfigPopup.svelte
```diff
@@ -0,0 +1,63 @@
...
+	<Content
+		style="display: flex; flex-direction: column; width: 400px; max-height: 80vh; overflow-y: scroll"
```
cabreraalex, line 46: here and below, can we use tailwind?

frontend/src/lib/components/popups/ChartConfigPopup.svelte
```diff
@@ -0,0 +1,63 @@
...
+	function undoChartConfig() {
+		config = oldConfig;
+		dispatch('close');
+	}
+
+	function resetChartConfig() {
+		zenoClient.deleteChartConfig(config.projectUuid, chartId).then(() => {
+			zenoClient.getChartConfig(config.projectUuid, chartId).then((fetchedConfig) => {
+				config = fetchedConfig;
+				dispatch('close');
```
cabreraalex, line 31: I don't think this should close on reset, might want to reset and then change them again.

PR #595, opened 2023-12-21
fix: correct diff comparison and add boolean comparison with three st…

Sparkier:
…ates

# Description

fix ZEN-428

For boolean columns this returns -1 if a is false and b is true, 0 if both are the same, and 1 if a is true and b is false.

This makes it possible to sort by model differences where one wants to find out if a better than b or b better than a.


neubig (approved):
Looks great, thanks!

PR #596, opened 2023-12-22
fix: gracefully fail charts if columns have changed

Sparkier:
# Description

fix ZEN-414

This disentangles the fetching of chart data from the chart. The reason behind this is that fetching chart data has some failure cases (e.g. a column for a slice filter does not exist anymore) that should not break charts altogether. It should still be possible to change the chart spec. Furthermore, this adds an error message if the chart data could not be loaded.


Sparkier:
Our openapi codegen creates the function for fetching the data twice for some reason and I don't really understand why.

cabreraalex:
yeah odd it did for me too

cabreraalex (approved):
LGTM, have to make sure it plays well with the new settings PR after merge.

PR #597, opened 2023-12-22
style: convert a lot of style tags to class tags with tailwind

Sparkier:


cabreraalex (approved):
Doin the lord's work i tell ya

PR #598, opened 2023-12-22
feat: unify header across views

cabreraalex:
# Description

Fix ZEN-429

image


cabreraalex:
Charts still looks off, but can fix in another PR where we add headers describing what view you are in probably.

Sparkier (changes requested):
Really like it overall! Some small things:
- Can we make the project/report name a bit bigger? Important to know and not visually prominent at all now. More like it was before.
- How's the link icon useful on project home? Don't think we need it there.
- One thing that was really nice before was that the project name started right with the metadata panel. Probably not possible if we want to keep the full logo, though, which I find nice.

cabreraalex:
Updated the font, agreed it looks nice.

I like the link button on projects b/c it removes the extra state for a clean URL I find useful. can go either way though.

Yeah I think worth having the consistent logo on all screens.

image

Sparkier (approved):
Like the button on projects, too. I was talking about home, though.

image

cabreraalex:
A couple of changes:
* Made settings a gear icon to be consistent
* fixed mobile view with long titles
* Fixed tests

image

Sparkier:
> A couple of changes:
> 
> * Made settings a gear icon to be consistent
> * fixed mobile view with long titles
> * Fixed tests
> 
> image

Gear looks off-center in your screenshot.

cabreraalex:
> Like the button on projects, too. I was talking about home, though.
> 
> image

oh LOL that should definitely go

Sparkier:
> > A couple of changes:
> > 
> > * Made settings a gear icon to be consistent
> > * fixed mobile view with long titles
> > * Fixed tests
> > 
> > image
> 
> Gear looks off-center in your screenshot.

Also would probably swap with question mark for consistency with home.

Sparkier:
Should we replace the explore button as well? Is the positioning even right for it?

cabreraalex:
> > > A couple of changes:
> > > 
> > > * Made settings a gear icon to be consistent
> > > * fixed mobile view with long titles
> > > * Fixed tests
> > > 
> > > image
> > 
> > 
> > Gear looks off-center in your screenshot.
> 
> Also would probably swap with question mark for consistency with home.

I switched them in both, always at the far left now so consistent everywhere

Sparkier:
Question mark left? One thing that I still find confusing about that is that the question mark is not related to the project as is the user icon whereas the gear is, but the gear is between the two. I think question mark next to user might make more sense?

cabreraalex:
hmm maybe far right? but user far right is most typical right

Sparkier:
Yes I think user makes sense. Why do you want to have the question mark on the left?

cabreraalex:
always same place. but I guess it's always same place next to the user icon or sign in stuff... can move it back I think you're right makes more sense

Sparkier:
Maybe the settings button should be close to the project content anyway. Like next to the heart or something.

cabreraalex:
Okay I moved the help to always be on the far left of the right side, made the most sense when using it. Also changed the cog icon

image

PR #612, opened 2023-12-27
feat: add warning if project visibility is narrower than report

Sparkier:
# Description

fix ZEN-431


cabreraalex (approved):
I suggest showing the warning inline with the project dropdown, wrap them in a div:

image

PR #614, opened 2023-12-28
fix: don't show a continuous histogram if there is only one value

Sparkier:


backend/zeno_backend/processing/histogram_processing.py
```diff
@@ -124,8 +124,13 @@ async def histogram_metric_and_count(
     else:
         calculate_histograms = False
 
+    # if end = start, can't show a bar, remove bucket (only happens if signular value)
```
cabreraalex, line 127: sp, singular

cabreraalex: also ==

cabreraalex: Does this not show the feature in the histogram then? is that what we want to happen? or do we want to just show the singular bar so you know what the single value is for everything?

cabreraalex (changes requested):

backend/zeno_backend/processing/histogram_processing.py
Sparkier: Previously the histogram was completely empty for some reason so I removed it.

Don't know if we want it single value histogram is not useful really but you see that there is a column... could go either way.

cabreraalex (approved):

PR #615, opened 2023-12-28
style: unify header buttons

Sparkier:
# Description

old: 
image

new:
image


cabreraalex:
We should maybe rethink how we show the tab between "my hub" and explore, make it more central and obvious it is a sort of tab? Not very discoverable now. but this is okay for now I think.

cabreraalex (approved):

Sparkier:
> We should maybe rethink how we show the tab between "my hub" and explore, make it more central and obvious it is a sort of tab? Not very discoverable now. but this is okay for now I think.

Agreed. Should think about positioning of these things. Now we have settings next to navigation kinda weird.

PR #616, opened 2023-12-28
feat: allow project editor to update data

Sparkier:
# Description

Enables project editors to update project data and systems.

fix ZEN-403

Requires https://github.com/zeno-ml/zeno-client/pull/50

### DIFF
--- a/backend/zeno_backend/routers/sdk.py
+++ b/backend/zeno_backend/routers/sdk.py
@@ -1,5 +1,6 @@
 """FastAPI server endpoints for the Zeno SDK."""
 import uuid
+from urllib import parse
 
 import pyarrow as pa
 from amplitude import BaseEvent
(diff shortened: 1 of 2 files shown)


backend/zeno_backend/routers/sdk.py
```diff
@@ -86,6 +88,34 @@ async def verify_api_key(self, api_key: str) -> bool:
...
+    Args:
+        owner_name (str): name of the project's owner.
+        project_name (str): name of the project.
+        api_key (_type_, optional): API key of the user making the request.
+            Defaults to Depends(APIKeyBearer()).
+
+    Returns:
+        str: uuid of the requested project.
+    """
+    uuid = await select.project_uuid(owner_name, parse.unquote(project_name))
```
cabreraalex, line 106: should we parse unquote the project name to be safe?

backend/zeno_backend/routers/sdk.py
```diff
@@ -86,6 +88,34 @@ async def verify_api_key(self, api_key: str) -> bool:
...
+    if uuid is None:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=(
+                "ERROR: Project " + owner_name + "/" + project_name + " does not exist."
```
cabreraalex, line 111: since we don't use this style maybe we say something like

"Project ____ for user ____ does not exist"

cabreraalex (approved):

backend/zeno_backend/routers/sdk.py
Sparkier: You mean the owner name?

PR #618, opened 2023-12-29
fix: add back actions for beeswarm chart

Sparkier:
# Description

fix ZEN-423


cabreraalex:
This is a bit odd because it shows for every beeswarm, but only exports the legend for the first one right? Not sure how to handle this...

Sparkier:
> This is a bit odd because it shows for every beeswarm, but only exports the legend for the first one right? Not sure how to handle this...

Oh good catch! Did not notice but that's probably why we removed it 😂

Let's leave this open for the new year then.

PR #631, opened 2024-01-04
fix: tooltip updates and loading spinner on navigate

cabreraalex:
# Description

Switching between explore and home wasn't working well. ALso added spinner to header whenever navigating to show if it's being slow after you clicked (necessary after removing view transitions)

### DIFF
--- a/frontend/src/lib/util/tooltip.ts
+++ b/frontend/src/lib/util/tooltip.ts
@@ -1,19 +1,17 @@
 import { tooltipState } from '$lib/stores';
 
 export function tooltip(node: HTMLElement, params: { text: string }) {
-	let isHovering = false;
-
 	const handleFocus = function (e: MouseEvent) {
-		isHovering = true;
 		tooltipState.set({
 			hover: true,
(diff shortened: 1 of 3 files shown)


frontend/src/lib/util/tooltip.ts
```diff
@@ -1,31 +1,43 @@
 import { tooltipState } from '$lib/stores';
 
 export function tooltip(node: HTMLElement, params: { text: string }) {
-	function handleFocus(e: MouseEvent) {
-		tooltipState.set({ hover: true, mousePos: { x: e.clientX, y: e.clientY }, text: params.text });
+	let isHovering = false;
```
Sparkier, line 4: Why this change? Before we were not adding listeners that are not needed. Now we call the move function even if we're not hovering. Seems unnecessary.

cabreraalex: Ah good point. Changed it back but still in a way that the tooltip text updates.

Sparkier (approved):

PR #632, opened 2024-01-06
fix: radar chart issues

Sparkier:
# Description

Fixes some sizing issues with radar charts.


cabreraalex:
looks like the package-lock thing breaks npm ci in CI

Sparkier:
> looks like the package-lock thing breaks npm ci in CI

Yep seems like. I think it somewhat broke docker for me though maybe worth testing some more at some point.

cabreraalex (approved):

PR #643, opened 2024-01-10
fix: show projects only on desktop devices

Sparkier:
# Description

fix ZEN-271

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -22,6 +22,7 @@
 				"isomorphic-dompurify": "^1.8.0",
 				"marked": "^11.1.1",
 				"svelte-codemirror-editor": "^1.1.0",
+				"svelte-device-info": "^1.0.0",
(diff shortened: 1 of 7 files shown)


cabreraalex:
Do we need another lib for this? ChatGPT says we can use something like this:

```
function isMobileDevice() {
    return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);
}
```

Sparkier:
I guess that's a philosophical question. This library is pretty tiny and does a bit more of a check than what ChatGPT told you but we could definitely just copy this part:

https://github.com/rozek/svelte-device-info/blob/517210b3ec8caa65606f066572cfb85a03841728/src/svelte-device-info.ts#L10-L15

Problem then is that we have to maintain this in case this list ever changes.

cabreraalex:
SGTM!

cabreraalex (approved):

PR #645, opened 2024-01-11
fix: only show user orgs to add to projects

### LINKED ISSUE #644: Auto-population of dropdown fields in settings
Collaborators and Organizations dropdown fields auto-populate in the settings pop-up of a project. These should be scoped to a given organization rather than what seems to be all of them.

image

System:
MacOS

Chrome

cabreraalex:
# Description

Fix #644


Sparkier:
I don't know if this is what we want here. E.g. in the student project we explicitly used this to be able to share with an org they are not a member of.

PR #766, opened 2025-01-04
fix: don't show ouptut in compare

Sparkier:
If output is selected, it does not really make sense to show it since the instance view is shown right below anyway.


neubig (approved):
LGTM, thanks! Although it looks like we'll need to fix an integration test

